PR #30, opened 2014-11-21
Add hash_equals for hmac comparision

0xbb:
Hi!
First, thanks for creating the crypto library!
I agree with your conclusions that 9/10 PHP crypto snippets on the web are insecure and people are better off with your library.

So I did some review of the ``VerifyHMAC`` function and noticed that the now built-in function ``hash_equals`` could be used.


sarciszewski:
:+1: to this, using built-in primitives when available is a wise move.

defuse:
Sorry for taking so long to review this (wow, November 2014).

The arguments to `hash_equals()` are in the wrong order. The one that's computed from the message should be the first parameter (`$known_string`) and the one included in the message (provided by the attacker) should be the second parameter (`$user_string`).

I don't have time to do it today, but I'll merge this into a different branch and then fix the order.

paragonie-scott:
I'm about to refactor the library to use PSR-4. This will probably create a merge conflict with this PR.

defuse:
That's okay, it's probably easier if I just copy/paste the code and give credit to @0xbb in the commit comment, than it would be to do the whole branch, change the parameter order, etc. stuff.

PR #33, opened 2015-01-16
Use openssl instead of mcrypt.

### LINKED ISSUE #32: Mcrypt is abandoned, use OpenSSL
Making a note for my own fork.

sarciszewski:
Resolves #32

### DIFF
--- a/Crypto.php
+++ b/Crypto.php
@@ -283,6 +283,7 @@ final class Crypto
         
         $method = self::CIPHER.'-'.self::CIPHER_MODE;
         
+        self::EnsureConstantExists("OPENSSL_RAW_DATA");
         self::EnsureFunctionExists("openssl_encrypt");
         $ciphertext = openssl_encrypt(
             $plaintext,
@@ -307,6 +308,7 @@ final class Crypto
         
         $method = self::CIPHER.'-'.self::CIPHER_MODE;
         
+        self::EnsureConstantExists("OPENSSL_RAW_DATA");
         self::EnsureFunctionExists("openssl_encrypt");
         $plaintext = openssl_decrypt(
             $ciphertext,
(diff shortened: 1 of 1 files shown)


narfbg:
@sarciszewski No immediate problems at first glance, but you should just set `CIPHER` to 'aes-128' instead of computing `$method` dynamically.

sarciszewski:
Yeah, I'll send an updated commit tonight :)

sarciszewski:
Test fails on PHP 5.3 which was EOL'd months ago due to an undefined constant (`OPENSSL_RAW_DATA`). http://php.net/eol.php
Test fails on PHP 5.2.

I recommend, in the spirit of #36 support be enforced for a minimum of PHP 5.4.

For users who are on 5.2/5.3, I recommend branching/tagging before merging so there's an mcrypt version lying around.

Crypto.php
```diff
@@ -304,42 +305,22 @@ private static function PlainEncrypt($plaintext, $key, $iv)
...
+            $method,
+            $key,
+            OPENSSL_RAW_DATA,
+            $iv
+        );
         if ($plaintext === FALSE) {
             throw new CannotPerformOperationException();
         }
-
+
```
defuse: Evil whitespace!

tsyr2ko:
*SecureRandom* function still uses mcrypt_* function.
Should use *openssl_random_pseudo_bytes* instead

sarciszewski:
Is `openssl_random_pseudo_bytes()` safe to use?

narfbg:
As long as you use its second parameter to check if a strong-enough source was used, I regard it as better than fopen()-ing /dev/urandom and trying to do it all yourself.

sarciszewski:
I meant from a cryptographer's perspective. I managed to get to `RAND_SSLeay()` in the openssl codebase before getting a headache.

PR #36, opened 2015-01-16
Remove Support for PHP 5.2 and PHP 5.3

sarciszewski:
Only supported PHP versions should be covered cc @ircmaxell :)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,8 @@ php:
     - "5.6"
     - "5.5"
     - "5.4"
+    - "5.3"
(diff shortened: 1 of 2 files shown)


defuse:
Apparently this didn't work because the build fails: https://travis-ci.org/defuse/php-encryption/jobs/48136266

defuse:
I fixed it, the versions of PHP in the allowed failures had to be quoted strings.

defuse:
Note that I opened #37 about actually checking that they fail safely.

PR #40, opened 2015-02-16
Use openssl_ for CSPRNG

sarciszewski:
Make `mcrypt_create_iv()` optional.

### DIFF
--- a/Crypto.php
+++ b/Crypto.php
@@ -342,12 +342,22 @@ final class Crypto
      */
     private static function SecureRandom($octets)
     {
-        self::EnsureFunctionExists("mcrypt_create_iv");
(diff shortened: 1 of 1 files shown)


defuse:
Note to self: We can remove the mcrypt dependency in composer after this.

defuse:
Should we keep the `mcrypt_create_iv` at all? If they don't have OpenSSL, the encryption won't work, right, or is there a circumstance where encryption will work but the openssl RNG isn't available?

sarciszewski:
Since we need 5.4.x now, we can get by with just openssl. And if openssl's PRNG isn't secure, TLS is doomed.

defuse:
Okay, let's get rid of the `mcrypt_create_iv()`

sarciszewski:
This should be ready to merge, then.

Issue #41, opened 2015-03-02
Making it easier for people to use...


mattsah:
@defuse

I came across your library in some of those slim threads and thought it might be great to use, as I've currently been investigating encryption libraries to finish my security module.

That said, two things I noticed immediately:

1) It lacks composer support
2) It takes over the exception handler (I understand this is for security)

I'd like to fix these two problems and make use of your library, but based on the previous composer.json pull request that is open, I'd like to verify that you're interested first.

With respect to the exception handler, it would make more sense, if the exception handler was changed temporarily during the processes which could throw exceptions.  Your exception handler could be split to a separate class, and http://php.net/manual/en/function.restore-exception-handler.php can be used to restore the previous one once decrypting is done (for example).

sarciszewski:
This is sort of a drop-in library. I could look into forking this to make it composable.

defuse:
I need to do some research to find out how composer works. 

I like the idea of hooking the exception handler only temporarily. I think it's better because it's less likely that the client will have overridden us and replaced our handler with a(nother) vulnerable one.

mattsah:
@sarciszewski 

There's a pull request to make it work with composer.

mattsah:
@defuse 

Please see https://github.com/defuse/php-encryption/pull/42

This removes the exception handler as a separate class and sets it as a simple callback function.  I added two methods to set and reset the exception handler and wrapped the Encrypt and Decrypt methods with them.

It was not clear to me wether or not the CreateNewRandomKey really needed it, I'll leave that up to your judgement.

I also did a bunch of docblock cleanup and generated some simple markdown docs for the API using sage (https://github.com/dotink/sage).

This combined with the composer pull request (which will need some modification since I moved Crypto to an src sub directory for doc purposes) should make this library fairly simple to use for other developers (whether that be the guys from Slim, or myself).

I ran the test script with passing output, and also did a quick modify in the example to test the exception handler was working as implemented (although i did not commit that change as it was an intentional fail).

pulyaevskiy:
@defuse Please see #31 which adds support for composer. I can update it to fit #42 (moving Crypto.php to `src/`).

defuse:
Hang on, I just realized the set-then-restore exception handler pattern won't work. If a crypto exception is correctly caught, the method call to restore the exception handler won't get run, and we'll have permanently changed it. :-(

mattsah:
@defuse Still better than right now, no?  Which is that the exception handler is taken over completely.

[...]

PR #31, opened 2014-11-23
Added composer.json

pulyaevskiy:
Hi defuse,

Thanks for this library!

This PR just adds very simple composer.json so that it's possible to include this library in project dependencies easily using composer.

PR #42, opened 2015-03-02
Made exception handler isolated on crypto functions only + cleanup

mattsah:
Cleanup was done in order to provide clearer API documentation for
other users which appeared to be lacking.  Documentation was
generated using sage (https://github.com/dotink/sage), hence the
addition of the sage.config

### DIFF
--- a/Crypto.php
+++ b/Crypto.php
@@ -90,10 +90,17 @@
  *  - The ciphertext is invalid or not in the correct format.
  *  - The attacker modified the ciphertext.
  */
-class InvalidCiphertextException extends Exception {}
+class CryptoException extends Exception {
+    public function __construct($message = NULL, $code = 0, Exception $previous = null) {
+        parent::__construct($message, $code, $previous);
+        restore_exception_handler();
+    }
+}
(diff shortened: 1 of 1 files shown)


mattsah:
This solves half of the issues in #41, the other half is in https://github.com/defuse/php-encryption/pull/31 although would need to be modified to load src/Crypto.php not just Crypto.php.

If both of these are pulled in, and you register the package on packagist, I think you'd see a lot more people using this which could only be a good thing all around.

defuse:
Thanks, I'll review this as soon as I have time.

Would it be easy to make separate pull requests for the documentation and exception handling (or at least have them in separate commits)? It would be nice to have security-critical changes isolated. If not (like, if it's not something that would take 2 minutes), don't worry about it.

mattsah:
I don't know if I can modify this pull request per se.  I could split them into two separate commit though (which may or may not modify this original pull request).

mattsah:
I'm closing this and moving to the split which also has a fix for the aforementioned issue.

PR #43, opened 2015-03-02
Use more environment friendly exception handling

mattsah:
This causes the handler to be registered with each function and then restored
when an exception is thrown or in the case that the method returns proper.

### DIFF
--- a/Crypto.php
+++ b/Crypto.php
@@ -90,10 +90,17 @@
  *  - The ciphertext is invalid or not in the correct format.
  *  - The attacker modified the ciphertext.
  */
-class InvalidCiphertextException extends Exception {}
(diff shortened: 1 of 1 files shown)


mattsah:
@defuse 

This is split and also fixes the problem you brought up on the previous one.  I closed the other request.

This one should enable restoring the exception handler in either case, and doesn't have all the doc cleanup.  Please let me know if you still want the docs and cleanup.

mattsah:
Please hold on this.  Additional testing has shown it *does not always work*.  Looking for something more reliable.

mattsah:
Closing this... after multiple additional attempts... this just doesn't seem viable.

defuse:
Thanks very much for trying!

Issue #50, opened 2015-03-14
Namespacing


defuse:
Include the crypto class and exception classes inside a namespace so we're less likely to conflict with other code.

defuse:
This is blocking adoption: https://github.com/Webiny/Framework/issues/39#issuecomment-80982903

sarciszewski:
Do you want to just use `namespace Defuse/Crypto`?

mathewpeterson:
I would like to suggest complying with PSR-4 instead of just adding a namespace: https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-4-autoloader.md

By adopting psr-4, it would also require separating out the symbols (classes, exceptions) into their own file.

sarciszewski:
PSR-4 entails a slight refactor. Feel free to assign this to me ;)

sarciszewski:
- Rule #4: The library should consist of a single PHP file and nothing more.

    > Some PHP encryption libraries, like libsodium-php [1], are not
    > straightforward to install and cannot packaged with "just download and
    > extract" applications. This library will always be just one PHP file
    > that you can put in your source tree and require().

PSR-4 is not compatible with this goal. PSR-4 compatible namespacing (`Defuse\Crypto\Crypto`) is though.

josephdpurcell:
@sarciszewski / @defuse, forgive my ignorance, but is there a security reason to not follow PSR standards?

The reason for goal #4 (keeping all classes in a single file) appears to be for ease of use. But, "ease" is not determined by what is easy for `require()` but what is easy for an autoloader. The trend in PHP is to use an autoloader and the composer implementation which follows PSR is already widely adopted.

sarciszewski:
There are a couple of ways we can do this:

1. Maintain two branches (e.g. `v.1` and `v.2`), the latter which is in PSR-4 format.
2. Fork the project to use PSR-4.
3. Forsake the original Rule 4, acknowledge that everyone is moving towards composer/PSR-4, get with the trends.

I'll leave this decision with @defuse, but I'm fine with either of the three.

anlutro:
You could use a script that concatenates all classes/functions + a simple bootstrap.php into a single PHP file and distribute that as a binary file with each Github release, or just put it in a dist directory.

defuse:
[...]

PR #57, opened 2015-05-16
PSR-4 Refactor (Thanks @0xbb for #30)

paragonie-scott:
Closes #30, #50 

This should probably be released as version `2.0.0` when it has been reviewed and merged.


defuse:
Good news: I learned to stop worrying and love the backslash.

defuse:
Only problems I see:

- Some of the functions in the global namespace are missing leading backslashes.
- That typo redragonx found (which isn't actually a problem with this PR, just noting it so I'll fix it later).
- The exception handler object doesn't get instantiated when you just autoload/include the stuff?

paragonie-scott:
Good catch. I'll look into getting composer to set that up and edit autoload.php as well. And I'll double check global namespace resolution too :)

defuse:
Thank you!

composer.json
```diff
@@ -10,7 +11,7 @@
         }
     ],
     "autoload": {
-        "files": ["Crypto.php"]
+        "files": ["src/autoload.php"]
```
mathewpeterson: If the library is PSR4 compliant then you should be able to use the ["psr4" autoloader](https://getcomposer.org/doc/01-basic-usage.md#autoloading).

autoload.php
```diff
@@ -0,0 +1,38 @@
+<?php
+/**
+ * PSR-4 compatible autoloader
+ */
+\spl_autoload_register(function ($class) {
```
mathewpeterson: Since the library only has ~5 files, it would just be more efficient to simply require each of them here. 

There's no need to reinvent the wheel.

composer.json
paragonie-scott: Well, it *is*, except `autoloader.php` also overrides the exception handler.

https://github.com/defuse/php-encryption/blob/61e88d6c8a29f295a82f7b22335a549f589df828/autoload.php#L38

autoload.php
paragonie-scott: I'm submitting a PR for a streaming API soon: https://github.com/paragonie-scott/php-crypto-stream

It will be a self-contained class in the same namespace.

composer.json
mathewpeterson: Yeah, that is an interesting approach. Attaching itself to exception handler during class instantiation is very odd. I read the comments in the file so I understand why it was done that way. 

What if you used composer psr4 for autoloading and then use this files directive for including the handler? It just feel wrong to write your own autoloader these days..

Issue #60, opened 2015-05-18
Best way to store the cypher key


bab55z:
What is the best way to store the key used to encrypt data?

I used base64_encode/decode my cypher key and then save it to a file, but this is not advised according to the crypto class comments.

Thus, I am asking what are or is the best way to store a cypher key to avoid leaks and attacks.

thanks.

sarciszewski:
There are none yet. I proposed a patch to PHP 7 but it got locked out due to a feature freeze, politics, and processes.

Note to self: create a PHP-land implementation of `Sodium::bin2hex()` and `Sodium::hex2bin()` to solve this problem once and for all.

pnowosie:
Maybe (#5) KeyLocker could help here a little? So we could protect encryption key with a password.

timoh6:
Consider storing the actual key as plain hex-encoded string into the file, and when needing it you hash it and use the raw output as the key for the cipher.

You can hash the hex string with SHA256 and use the raw output as a key (or truncate it to 128 bits for AES-128).

There are no timing leaks with SHA256 hashing.

(this assumes the key is strong and doesn't need stretching)

pnowosie:
@timoh6, He is already storing the key in the file plaintext. Hex/Base64 encoding does not add security to key storage, and I'm afraid hashing the key to derive another one is pointless. If I got @bab55z correctly, he want to protect the stored key from leaking.

timoh6:
@pnowosie The reason to use a hash function is to turn the key into an acceptable encryption key, _without_ needing to perform dubious string decondings (like base64_decode).

To me the question is indeed asking how he can avoid the side-channel leak created by base64_decode (unless I'm missin gsomething).

pnowosie:
Oh, I missed that text-to-byte decoding could leak information via side-channel, good point @timoh6.
So in this case, key material stored in file can be anything (secure random) and much longer than actual 128 bit key. But still not secure in case of storage...

bab55z:
it seems like there isn't yet a 100% sure way to safely store the cypher key and avoid leakings!

sarciszewski:
As soon as PR #57 is reviewed and merged I will be contributing a constant
time bin2hex and hex2bin to this project and encourage its use :)

spaze:
I "encode" my keys as PHP strings using the PHP's `\x[0-9A-Fa-f]{1,2}` format so they look like

```
[...]

PR #61, opened 2015-06-11
Contant-Time bin2hex/hex2bin -- closes #18 #60

### LINKED ISSUE #18: Encoding
We need to provide encoding and decoding functions (hex, base64, etc) that are safe against side channel attacks.

paragonie-scott:
[Referenced implementation](https://github.com/jedisct1/libsodium/blob/b2dd3453d57dc3a98b9704e3fd1326483ed4e47b/src/libsodium/sodium/utils.c#L99-L180).

### DIFF
--- a/src/Crypto.php
+++ b/src/Crypto.php
@@ -712,7 +712,9 @@ final class Crypto
             $c_alpha = ($c & ~32) - 55;
             $c_alpha0 = (($c_alpha - 10) ^ ($c_alpha - 16)) >> 8;
             if (($c_num0 | $c_alpha0) === 0) {
-                break;
+                throw new \DomainException(
+                    'Crypto::hexToBin() only expects hexadecimal characters'
+                );
             }
             $c_val = ($c_num0 & $c_num) | ($c_alpha & $c_alpha0);
             if ($state === 0) {
--- a/src/Crypto.php
+++ b/src/Crypto.php
@@ -671,4 +671,58 @@ final class Crypto
             return \substr($str, $start);
         }
     }
+    /**
+     * Convert a binary string into a hexadecimal string without cache-timing 
+     * leaks
+     * 
+     * @param string $bin_string (raw binary)
+     * @return string
+     */
+    public static function binToHex($bin_string)
+    {
+        $hex = '';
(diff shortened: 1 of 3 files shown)


src/Crypto.php
```diff
@@ -671,4 +671,58 @@ private static function ourSubstr($str, $start, $length = null)
...
+        $state = 0;
+        
+        while ($hex_pos < $hex_len) {
+            $c = \ord($hex_string[$hex_pos]);
+            $c_num = $c ^ 48;
+            $c_num0 = ($c_num - 10) >> 8;
+            $c_alpha = ($c & ~32) - 55;
+            $c_alpha0 = (($c_alpha - 10) ^ ($c_alpha - 16)) >> 8;
+            if (($c_num0 | $c_alpha0) === 0) {
+                break;
```
defuse: This could lead to returning shortened strings which could have a security impact if the result is used as a key.

For example, they might accidentally have a space in their key "a81397 8912affa838387afffbc3278" and the resulting decoded key would be weak.

Better to be loud and throw an exception, is the built-in DomainException suitable?

paragonie-scott: That makes sense to me. :)

src/Crypto.php
```diff
@@ -671,4 +671,60 @@ private static function ourSubstr($str, $start, $length = null)
...
+            if (($c_num0 | $c_alpha0) === 0) {
+                throw new \DomainException(
+                    'Crypto::hexToBin() only expects hexadecimal characters'
+                );
+            }
+            $c_val = ($c_num0 & $c_num) | ($c_alpha & $c_alpha0);
+            if ($state === 0) {
+                $c_acc = $c_val * 16;
+            } else {
+                $bin .= \chr($c_acc | $c_val);
```
mtotheikle: Is it intentional that $c_acc is undefined in this case and will cause a PHP notice?

@defuse @paragonie-scott

paragonie-scott: $state is equal to 0 on the first iteration, so this notice will never be triggered. https://github.com/defuse/php-encryption/pull/77 defines it up top just to be semantically correct.

mtotheikle: Great!

Issue #38, opened 2015-02-06
Decryption fails with a lot of data. 


CapnKlutch:
I am trying to Encrypt/Decrypt large amounts of data. I am encrypting about 3.56MB of text, and when I  try to decrypt it, it fails on line 232 - ```throw new InvalidCiphertextException();```

My end goal is to be able to encrypt very large files perhaps up to 2gb, but at this point I'm stuck at decrypting 3.56MB.

sarciszewski:
Could you update to the latest version? There is no exception on line 213 for the current code.

CapnKlutch:
@sarciszewski My apologies it is on line ```232```

sarciszewski:
https://github.com/defuse/php-encryption/blob/master/Crypto.php#L197

HMAC verification is failing. This is peculiar. Could you upload a PoC with dummy/garbage data about how you are attempting to encrypt your data so we can debug this?

CapnKlutch:
@sarciszewski Here is the dummy data - http://tempsend.com/0EA469F46B

CapnKlutch:
@sarciszewski I also need to note that my ```const KEY_BYTE_SIZE = 32;``` from the 16 that shipped with it.

sarciszewski:
This works for me.
```
scott@vbox /mnt/share $ php -v
PHP 5.5.9-1ubuntu4.5 (cli) (built: Oct 29 2014 11:59:42) 
Copyright (c) 1997-2014 The PHP Group
Zend Engine v2.5.0, Copyright (c) 1998-2014 Zend Technologies
    with Zend OPcache v7.0.3, Copyright (c) 1999-2014, by Zend Technologies
scott@vbox /mnt/share $ cat poc.php 
<?php
include "Crypto.php";
try {
    $key = Crypto::CreateNewRandomKey();
    // WARNING: Do NOT encode $key with bin2hex() or base64_encode(),
    // they may leak the key to the attacker through side channels.
} catch (CryptoTestFailedException $ex) {
    die('Cannot safely create a key');
} catch (CannotPerformOperationException $ex) {
    die('Cannot safely create a key');
}

$message = file_get_contents('large.txt');

  try {
      $ciphertext = Crypto::Encrypt($message, $key);
  } catch (CryptoTestFailedException $ex) {
      die('Cannot safely perform encryption');
  } catch (CannotPerformOperationException $ex) {
      die('Cannot safely perform decryption');
  }

  try {
[...]

Issue #34, opened 2015-01-16
Version Tagging


sarciszewski:
This is a nice-to-have in case anyone wants to encrypt with version 1.x of this library in 2015 then decrypt it with version 3.x in 2025.

defuse:
+1, there should definitely be version tagging (even if all we do right now is check that it's equal to a constant).

sarciszewski:
So, how would we put it in here? In my own toy implementation, ciphertexts looked like this:

`"A1:[base64IV]:[base64ciphertext]:[base64hmac]"`

paragonie-scott:
I'd like to propose a migration path.

1. Make sure #63 is good, merge it.
2. Copy `Crypto.php` into `Legacy.php`
3. Add version tagging to `Crypto.php`
4. Add version tagging to `File.php`
5. Make sure the documentation (see #52) explains the migration path.

The final result should be released as `2.0.0`.

Thoughts?

defuse:
This is done.

DerMika:
Hi, I have a question.

I've started implementing an application which uses this encryption library to store encrypted values in a database. 

I have added a ~1.2.0 dependency to your library in Composer, but of course I do not want to fall behind on encryption. 

Will I run into problems with my encrypted data when I upgrade to the 2.0.0 release, or will there be a way to migrate my data when upgrading? 

For an application that will go live in one or two months time, what do you suggest I do: work with the 1.2 release and migrate later, or should I rather use the v2.0 branch in my project to avoid problems in the future?

PR #63, opened 2015-06-11
Large file encryption/decryption

### LINKED ISSUE #39: Support API for Encrypting/Decrypting files.
You can't encrypt super-large files with this library because you have to load them into RAM. We could either (1) Have a "stream" sort of API, where you feed it data and it feeds back ciphertext, but there are problems with this, e.g. when decrypting you'll get all of the plaintext and then only the last call will check the MAC, but the damage has already been done, or (2) Support file encryption/decryption only, so that we can encapsulate all of that.

paragonie-scott:
Differences from `Crypto`:
* AES-CTR instead of AES-CBC
* Operates on resources rather than strings
* Decryption is slightly slow; it recalculates the MAC over the entire file and doesn't decrypt if it fails to validate.

Changes to `Crypto`:
* All common denominators between the two classes have been moved to a separate class called `Core`.

Things we would like to test for before merging:
* Is nonce reuse sufficiently mitigated?
* Can we make the buffer size irrelevant to the decryption process?
* Should we implement versioning in the regular library **before** adding this feature, and then make the ciphertext here versioned too?

This should close #39 (and #38?) when it's merged.


pnowosie:
What's the point to moving to AES-CTR?

paragonie-scott:
OpenSSL with CBC mode adds PKCS7 padding to each chunk (for us, 8192 bytes), which adds 16 bytes of overhead to each chunk.

We also have to transform the IV after each chunk and we'd end up making a sort of counter anyway.

src/File.php
```diff
@@ -0,0 +1,397 @@
...
+        }
+        if (!is_string($outputFilename)) {
+            throw new Ex\InvalidInput("Output filename must be a string!");
+        }
+        
+        /**
+         * Open the file handles
+         */
+        // Input file handle
+        $if = \fopen($inputFilename, 'rb');
```
defuse: might return FALSE

src/File.php
```diff
@@ -0,0 +1,397 @@
...
+        \stream_set_read_buffer($if, self::BUFFER);
+        
+        // Output file handle
+        $of = \fopen($outputFilename, 'wb');
```
defuse: might return FALSE

src/File.php
```diff
@@ -0,0 +1,397 @@
...
+        \stream_set_write_buffer($of, 0);
+        
+        // Use encryptResource() to actually write the encrypted data to $of
+        $encrypted = self::encryptResource($if, $of, $key);
+        
+        // Close handlers
+        \fclose($if);
+        \fclose($of);
```
defuse: either of these could be fclosing FALSE (see the annotations above)

src/File.php
```diff
@@ -0,0 +1,397 @@
...
+        // Output file handle
+        $of = \fopen($outputFilename, 'wb');
+        \stream_set_write_buffer($of, 0);
+        
+        // Use decryptResource() to actually write the decrypted data to $of
+        $decrypted = self::decryptResource($if, $of, $key);
+        
+        // Close handlers
+        \fclose($if);
+        \fclose($of);
```
defuse: Same thing, explicitly check for fopen() failure

src/File.php
```diff
@@ -0,0 +1,397 @@
...
+     */
+    public static function encryptResource($inputHandle, $outputHandle, $key)
+    {
+        // Because we don't have strict typing in PHP 5
+        if (!is_resource($inputHandle)) {
+            throw new Ex\InvalidInput("Input handle must be a resource!");
+        }
+        if (!is_resource($outputHandle)) {
+            throw new Ex\InvalidInput("Output handle must be a resource!");
+        }
```
defuse: +1 for these type checks!

src/File.php
```diff
@@ -0,0 +1,397 @@
...
+        
+        // Sanity check; key must be the appropriate length!
+        if (self::ourStrlen($key) !== self::KEY_BYTE_SIZE) {
+            throw new Ex\InvalidInput("Bad key.");
```
defuse: Maybe more informative: "Incorrect key length." ?

src/File.php
```diff
@@ -0,0 +1,397 @@
...
+        $akey = self::HKDF(
+            self::HASH_FUNCTION,
+            $key,
+            self::KEY_BYTE_SIZE,
+            self::AUTHENTICATION_INFO
+        );
+        
+        /**
+         *  Generate a random initialization vector.
+         */
```
defuse: The mixing of multiline (/* */) comments and // comments is weirding me out here haha

src/File.php
```diff
@@ -0,0 +1,397 @@
...
+        $ivsize = openssl_cipher_iv_length($method);
+        if ($ivsize === FALSE || $ivsize <= 0) {
+            throw new Ex\CannotPerformOperation('Improper IV size');
+        }
+        $iv = self::secureRandom($ivsize);
+        
+        /**
+         * First let's write our IV to the first N blocks of the output file
+         */
+        \fwrite($outputHandle, $iv, $ivsize);
```
defuse: fwrite() can return FALSE, should be checked.

src/File.php
```diff
@@ -0,0 +1,397 @@
...
+        
+        /**
+         * We're going to initialize a HMAC-SHA256 with the given $akey
+         * and update it with each ciphertext chunk
+         */
+        $hmac = \hash_init(self::HASH_FUNCTION, HASH_HMAC, $akey);
```
defuse: I wonder if this can fail? The PHP docs don't say anything, but what happens if we passed "thishashfunctiondoesnotexist" as the first argument?

src/File.php
```diff
@@ -0,0 +1,397 @@
...
+        
+        /**
+         * Let's MAC our IV/nonce
+         */
+        \hash_update($hmac, $iv);
+        
+        /**
+         * Iterate until we reach the end of the input file
+         */
+        while (!\feof($inputHandle)) {
```
defuse: Note: feof will return true upon error as well.

src/File.php
```diff
@@ -0,0 +1,397 @@
...
+            $read = \fread($inputHandle, self::BUFFER);
```
defuse: Might return FALSE

src/File.php
```diff
@@ -0,0 +1,397 @@
...
+    
+    /**
+     * Increment a counter (prevent nonce reuse)
+     * 
+     * @param string $ctr - raw binary
+     * @param int $inc - how much?
+     * 
+     * @return string (raw binary)
+     */
+    protected static function incrementCounter($ctr, $inc = 1)
```
defuse: The default value is not used, I don't think? (better for an error to show if it's not given)

src/File.php
```diff
@@ -0,0 +1,397 @@
...
+    {
+        static $ivsize = null;
+        if ($ivsize === null) {
+            $ivsize = openssl_cipher_iv_length(self::CIPHER.'-'.self::CIPHER_MODE);
+        }
```
defuse: Should we just use `ourStrlen($ctr)`? There might be some weird multithreading stuff going on with different ciphers or something, in the future, that would break this thing. At the very least if we go this route we should check that `$ivsize` is the same as the strlen.

src/File.php
```diff
@@ -0,0 +1,397 @@
...
+        
+        // We start at the rightmost byte
+        for ($i = $ivsize - 1; $i >= 0; --$i) {
+            $c = \ord($ctr[$i]);
+            
+            $ctr[$i] = \chr(($c + $inc) & 0xFF);
```
defuse: If this `+` overflows the integer it'll get cast to float, we'll have problems. This isn't actually a problem since we're using a limited buffer size, but maybe we should put a bounds check on `$inc` just before this loop?

src/File.php
```diff
@@ -0,0 +1,397 @@
...
+            if (($c + $inc) < 255) {
```
defuse: That should be <=. A better way to do this is:

    $inc = ($inc >> 8) & (PHP_INT_MAX - 1);
    if ($inc === 0) { break; }

src/File.php
```diff
@@ -0,0 +1,397 @@
...
+                // We don't need to keep incrementing to the left unless we exceed 255
+                break;
+            }
+            $inc >>= 8;
```
defuse: This might sign-extend, so we should have a check above to make sure `$int` is not negative.

src/File.php
```diff
@@ -0,0 +1,397 @@
...
+         * Iterate until we reach the end of the input file
+         */
+        while (!\feof($inputHandle)) {
+            $read = \fread($inputHandle, self::BUFFER);
+            
+            // We use a PRF for successive IVs
+            $thisIv = self::incrementCounter($thisIv, $inc);
+            
+            /**
+             * Perform the AES encryption. Pads the plaintext in PKCS7 and then
```
defuse: Does this actually PKCS7 even though CTR mode?

src/File.php
```diff
@@ -0,0 +1,397 @@
...
+             * encrypts it.
+             */
+            $encrypted = \openssl_encrypt(
+                $read,
+                $method,
+                $ekey,
+                OPENSSL_RAW_DATA,
+                $thisIv
+            );
```
defuse: This can return FALSE as well.

src/File.php
paragonie-scott: No, that's one of the attractive features of OpenSSL's CTR implementation.

src/File.php
```diff
@@ -0,0 +1,397 @@
...
+            
+            /**
+             * Write the ciphertext to the output file
+             */
+            \fwrite($outputHandle, $encrypted, self::ourStrlen($encrypted));
```
defuse: Can return FALSE

src/File.php
```diff
@@ -0,0 +1,397 @@
...
+            /**
+             * Perform the AES encryption. Pads the plaintext in PKCS7 and then
+             * encrypts it.
+             */
+            $encrypted = \openssl_encrypt(
+                $read,
+                $method,
+                $ekey,
+                OPENSSL_RAW_DATA,
+                $thisIv
```
defuse: We need to be absolutely sure `openssl_encrypt` increments the IV internally in the same way we're incrementing our IV. What would the security consequences be, if, say, it started to increment little-endian instead of big-endian?

src/File.php
```diff
@@ -0,0 +1,397 @@
...
+                \hash_update($hmac, $read);
+            }
+            // We should now have enough data to generate an identical HMAC
+            $finalHMAC = \hash_final($hmac, true);
+        /**
+         * 3. Did we match?
+         */
+            if (!self::hashEquals($finalHMAC, $stored_mac)) {
+                throw new Ex\InvalidCiphertext();
+            }
```
defuse: TOCTTOU: Is there any way the file could now change at this point? For example, what if:

- The file is on Linux?
- The file is on Mac?
- The file is on a Windows NTFS or FAT?
- The file is on a Windows network share?
- The file is in Dropbox?
- etc.

src/File.php
```diff
@@ -0,0 +1,397 @@
...
+            $stored_mac = \fread($inputHandle, self::MAC_BYTE_SIZE);
+
+            // We begin recalculating the HMAC for the entire file...
+            $hmac = \hash_init(self::HASH_FUNCTION, HASH_HMAC, $akey);
+            
+            // Reset file pointer to the beginning of the file.
+            \fseek($inputHandle, 0, SEEK_SET);
+
+            // Set it to the first non-IV byte
+            \fseek($inputHandle, $ivsize, SEEK_CUR);
```
defuse: The above fseek()s and file ops can probably fail

src/File.php
```diff
@@ -0,0 +1,397 @@
...
+                /**
+                 * Would a full DBUFFER read put it past the end of the
+                 * ciphertext? If so, only return a portion of the file.
+                 */
+                if ($pos + self::BUFFER >= $cipher_end) {
+                    $breakW = true;
+                    $read = \fread($inputHandle, $cipher_end - $pos + 1);
+                } else {
+                    $read = \fread($inputHandle, self::BUFFER);
+                }
```
defuse: The above two freads() can fail. (And now it really matters because if they do, we truncate)

src/File.php
```diff
@@ -0,0 +1,397 @@
...
+                    self::ourStrlen($decrypted)
+                );
+                // Check result
+                if ($result === false) {
+                    throw new Ex\CannotPerformOperation(
+                        'Could not write to output file durind decryption.'
+                    );
+                }
+            }
+        return $result;
```
defuse: `$result` may not have been created yet if the loop never executes or `break;`s before the last bit, but it always executes once and doesn't `break;` so that's not a problem.

src/File.php
```diff
@@ -0,0 +1,397 @@
...
+                 */
+                $decrypted = \openssl_decrypt(
+                    $read,
+                    $method,
+                    $ekey,
+                    OPENSSL_RAW_DATA,
+                    $thisIv
+                );
+                
+                // Write the plaintext out to the
```
defuse: I think there's supposed to be a word after 'the'

src/Crypto.php
```diff
@@ -194,7 +188,7 @@ public static function decrypt($ciphertext, $key)
 
     /*
      * Runs tests.
-     * Raises CannotPerformOperationException or CryptoTestFailedException if
+     * Raises CannotPerformOperationException or Ex\CryptoTestFailed if
```
defuse: Does `Ex\` mean the same thing to us as the person reading this comment (i.e. do we know they made the same alias?)?

src/Core.php
```diff
@@ -0,0 +1,332 @@
+<?php
```
defuse: I'm assuming this file is 100% copypasted and therefore doesn't need any review. Correct me if I'm wrong.

paragonie-scott: Yep, it's cut from Crypto.php

src/File.php
```diff
@@ -0,0 +1,397 @@
...
+
+            /**
+             * Let's grab our MAC
+             * 
+             * It should be the last N blocks of the file (N = 32)
+             */
+            \fseek($inputHandle, (-1 * self::MAC_BYTE_SIZE), SEEK_END);
+            
+            // Grab our last position of ciphertext before we read the MAC
+            $cipher_end = \ftell($inputHandle) - 1;
```
defuse: I'm worried about 32-bit PHP and files > 2GB. Wouldn't that overflow the integer and give us something that's potentially non-exact?

src/File.php
paragonie-scott: (Before I squashed the commit, the line below did not read `if ($read === false)`...)

src/File.php
paragonie-scott: Should we error on file sizes >= `(PHP_INT_MAX - 1) >> 1`?

paragonie-scott:
Okay, that commit satisfied the code warts you spotted (thanks @defuse!).

* I still need to test `incrementCounter()` thoroughly to make sure it's doing The Right Thing (TM). *Momma says nonce reuse is the devil.*
* If I can guarantee it works properly, I need to find out why changing the buffer size causes decryption to fail after the first buffer chunk. Then we can make it irrelevant or configurable.
* We probably want to add version tagging to both libraries then rebase first. #34 

If @zooko has any spare cycles to tell me, quickly, if I'm *doing it wrong*, that would be greatly appreciated too! (For quick reference, this is a "read the whole file, verify MAC, then decrypt" approach, so no random access within the file -- maybe a separate class can be made for that?)

zooko:
I'm sorry! I can't look at this right now.

Regards,

Zooko

src/File.php
dchest: What about saving intermediate HMACs instead of hashing again? Use [hash_copy](https://php.net/manual/en/function.hash-copy.php) to copy the context, finalize and save the chunk's HMAC so far, and continue with the original?

paragonie-scott: I have no objections, although anything per-chunk means that the size of the buffer will matter a lot. This just means it will need to be stored in the metadata or configured based on the version of the library that was used to encrypt a file. (I was *trying* to see if we could make that detail irrelevant, but it never worked, so I'm not really that attached to this idea.)

defuse: @paragonie-scott The size of the buffer is still irrelevant. For example, you could have one implementation that buffers 1MiB, and another 10KiB, and they would both be compatible with each other. All that matters is that *within one decryption operation* you use the same buffer size for both passes.

paragonie-scott: Oh, I didn't see the edits to your comment!

I thought @dchest was saying to HMAC every chunk. I stand corrected.

dchest: @paragonie-scott yeah, I was suggesting a possible improvement to this part of @defuse 's comment:

> (MAC, SHA256 save, SHA256 check, decrypt)

Instead of saving SHA256, save intermediate MACs, then when decrypting, check intermediate MACs again.

paragonie-scott: The latest commit should have implemented this properly. I need to add a check that `$macs` is not empty, because if it is, it will raise an `E_NOTICE` error (yet still return `false`).

src/File.php
paragonie-scott: Both openssl and our implementation are big-endian.

src/File.php
paragonie-scott: The most likely error condition should be `fopen()` returning false, in which case we will not end up here.

defuse: I/O can fail at any time, for example if the disk got yanked out halfway through processing the file, or possibly if the adversary controlled the file permissions and changed them at just the right time. If it's an error rather than actually reaching the end of the file, we should raise an exception.

paragonie-scott: What would you recommend as a strategy? Grab `filesize()` at the beginning and do a sanity check that we've reached the finish line?

defuse: Hmm... The standard way to do it is to check `ferror()`, but I see PHP doesn't have such a function. Strange.

src/File.php
paragonie-scott: (This has mostly been reverted to use multiline (`/* */`) comments.)

src/File.php
paragonie-scott: (For people scanning the remarks, the commits since @defuse posted this one have added a check here on the following lines, but github doesn't mark this remark as outdated.)

paragonie-scott:
As per [this comment in r/netsec](https://www.reddit.com/r/netsec/comments/3akyy2/large_file_encryptiondecryption/csdmiua) I've added some basic documentation to this PR. I migrated the old documentation header from `Crypto.php` and wrote something similar for `File.php`.

Storing MACs in an array and re-verifying them on decryption (after the final HMAC has been verified) has been implemented, so race conditions against the filesystem should be effectively mitigated.

paragonie-scott:
Should I resubmit or just force push to my branch?

paragonie-scott:
Closing in favor of #78

PR #66, opened 2015-06-20
Ensure that openssl_decrypt is available instead of openssl_encrypt

Darkheir:
In the `plainDecrypt` method we actually are ensuring that the method `openssl_encrypt` is available but what we use later is `openssl_decrypt`.


paragonie-scott:
Good catch! :+1: In practice, if one is available, the other will be too, but this is definitely a more sane thing to check for.

PR #69, opened 2015-07-08
Fix a couple of spelling errors

chriseskow:


paragonie-scott:
:+1: It turned out this does not conflict with #63 at all, even though I thought it did. (Please ignore the previous commit reference. I've removed it.)

Issue #37, opened 2015-01-24
Safe Failure Tests on 5.2 and 5.3.


defuse:
Switching to OpenSSL broke 5.2 and 5.3. We don't want to support those versions any more, but we should at least be sure we're not *dangerous* on those versions. Add travis tests for those versions that make sure they fail somehow safe (e.g. by throwing an exception).

https://github.com/defuse/php-encryption/pull/36

sarciszewski:
I think this one should be closed now too.

defuse:
In #74, we removed the 5.2 and 5.3 tests. Re-add them with tests that check that the library isn't doing something dangerous (i.e. it should always throw an exception) on those versions.

defuse:
This is like worrying if the color of your bike shed will make it more likely to get broken into. Closing.

PR #74, opened 2015-07-19
remove unsupported php version from travis.yml

sstok:
Some minor improvements for Travis.

< PHP 5.4 is not supported so no need to test this, and use sudo to speed-up the boot process.

*Please ignore the typo in the branch-name...*

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,22 +1,13 @@
 language: php
 php:
+    - "7.0"
     - "5.6"
     - "5.5"
     - "5.4"
-    - "5.3"
-    - "5.2"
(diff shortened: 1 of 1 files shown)


defuse:
I'd like to eventually test that the library does not fail catastrophically on 5.2 and 5.3 (see #37). That's why the tests are still there, but allowed to fail. Running them now is a waste of time, so I'll remove them for the time being and make a note in #37 that they need to be re-added.

sstok:
PHP 5.2 isn't going to work as you get parsing errors :) `\substr` is invalid for PHP 5.2

PR #77, opened 2015-07-21
Make sure variable $c_acc is defined

paragonie-scott:


paragonie-scott:
Added this line to my file PR

Issue #71, opened 2015-07-17
Proposed Version 2.0.0 Roadmap


paragonie-scott:
After #63 has been reviewed and is merged, we should do #34. My proposal for #34 is as follows:

1. The first N bytes will be reserved for a version tag header. (N = 4, for example.)
2. For `Crypto.php`, we should add a method called `legacyDecrypt()` to retrieve the plaintext of messages encrypted with the 1.x strategy. File encryption/decryption should always be tagged as of the 2.0.0 release, so a backwards-compatible strategy is unnecessary. We should not allow legacy encryption.

------------

That would solve the thorniest issue, from my perspective.

Afterwards, I would say that the following issues should be part of the 2.0.0 milestone:

* #5 - We can use `hash_pbkdf2()` with sane constants and provide a compatible polyfill for PHP 5.4 (see [php-future](https://github.com/sarciszewski/php-future/blob/bd6c91fb924b2b35a3e4f4074a642868bd051baf/src/Security.php#L52-L87) for example)
* #68 - When this library is stable, we can use `random_bytes()` in PHP 7 and a polyfill that follows best practices on PHP 5.x, with no added conditional statements.
* ~~#70 - Will make it better to write `try/catch` blocks.~~

paragonie-scott:
* My Version 2.0.0 pull request is in #78 and should close the following tickets (not automatically however):
    * #34 
    * #39 
    * #76 

If my propsal #80 is accepted, we can encapsulate #5 in that functionality and close that one too. At that point, version 2.0.0 will require an instance of `\Defuse\Crypto\MasterKey`.

I might make this key piece a separate PR once 78 has been reviewed.

defuse:
Closing this because it doesn't have a well-defined goal, and I just made a milestone for v2.0 which we can use to track things.

PR #78, opened 2015-07-22
Proposed version 2.0 - file streaming, version tagging

### LINKED ISSUE #34: Version Tagging

### LINKED ISSUE #39: Support API for Encrypting/Decrypting files.

paragonie-scott:
Satisfies the proposal outlined in #71 

Supersedes #63 
Closes #34 
Closes #39


src/Crypto.php
```diff
@@ -62,10 +45,10 @@
      */
     public static function createNewRandomKey()
     {
-        self::runtimeTest();
-        return self::secureRandom(self::KEY_BYTE_SIZE);
+        $config = self::getVersionConfig(self::VERSION);
+        return self::secureRandom($config['KEY_BYTE_SIZE']);
     }
-
+
```
sstok: Trailing whitespace ;)

sstok:
https://github.com/paragonie/php-encryption/blob/files/src/Crypto.php#L322
`if (\in_array(self::CIPHER_METHOD, \openssl_get_cipher_methods()) === FALSE) {` was this left unchanged intentionally?

src/Crypto.php
```diff
@@ -253,22 +353,30 @@ public static function runtimeTest()
...
             $plaintext,
-            self::CIPHER_METHOD,
+            $config['CIPHER_METHOD'],
             $key,
             OPENSSL_RAW_DATA,
             $iv
         );
 
         if ($ciphertext === false) {
+            \var_dump([
```
sstok: Debug code

src/Crypto.php
```diff
@@ -409,39 +432,21 @@ private static function HKDF($hash, $ikm, $length, $info = '', $salt = null)
...
-        $blind = self::createNewRandomKey();
-        $message_compare = \hash_hmac(self::HASH_FUNCTION, $message_hmac, $blind);
-        $correct_compare = \hash_hmac(self::HASH_FUNCTION, $correct_hmac, $blind);
-        return $correct_compare === $message_compare;
+        $message_hmac = \hash_hmac($config['HASH_FUNCTION'], $message, $key, true);
+        return self::hashEquals($correct_hmac, $message_hmac);
     }
-
-    private static function testEncryptDecrypt()
+
```
sstok: Trailing whitespace ;)

src/Crypto.php
```diff
@@ -516,13 +524,18 @@ private static function HKDFTestVector()
...
-        $okm = self::hexToBytes(
+        $okm = self::hexToBin(
             "2c91117204d745f3500d636a62f64f0a" .
             "b3bae548aa53d423b0d1f27ebba6f5e5" .
             "673a081d70cce7acfc48"
         );
-        $computed_okm = self::HKDF("sha1", $ikm, $length);
+        $computed_okm = self::HKDF("sha1", $ikm, $length, '', null, $config);
         if ($computed_okm !== $okm) {
+            \var_dump([
```
sstok: Debug code

src/File.php
```diff
@@ -0,0 +1,762 @@
...
+             */
+            if(\fseek($inputHandle, (-1 * $config['MAC_BYTE_SIZE']), SEEK_END) === false) {
+                throw new Ex\CannotPerformOperation(
+                    'Cannot seek to beginning of MAC within input file'
+                );
+            }
+            
+            // Grab our last position of ciphertext before we read the MAC
+            $cipher_end = \ftell($inputHandle) - 1;
+            if ($cipher_end === false) {
```
sstok: This is going to fail as with the `-1` its converted to an integer which doesn't strict equal false!

tests/stream/decrypt.php
```diff
@@ -0,0 +1,43 @@
+<?php
+require_once \dirname(\dirname(__DIR__)).'/autoload.php';
+
+if (!\file_exists('key.txt')) {
+    echo 'You need to generate an key first!', "\n";
```
sstok: *a* key (not *an*) :)

paragonie-scott:
Ugh, wow, I can't believe I forgot to remove my debug code. Stupid, stupid, stupid.

Commit incoming.

paragonie-scott:
The latest test verifies that our nonce incrementing strategy is functioning correctly. Aside from the danger of birthday collisions, nonce collisions are mitigated.

Mitigating nonce reuse is nontrivial, since we're limited to 128 bits. I think we would first need to create a subkey for each file and store the random HKDF or PBKDF2 salt in the file data to make this impractical.

paragonie-scott:
Since HKDF accepts a salt parameter, I added a random per-file salt to the file encrypted file header in addition to the IV.

Without the salt, the odds of a birthday collision after 2^64 files is approximately 50%. Depending on the file size, a nonce reuse becomes a real problem.

Using a random salt and a nonrandom nonce (e.g. all 0's) is virtually identical to the existing situation.

Using a random salt AND a random nonce raises our birthday resistance to require 2^128 files for a 50% chance of birthday collision. (This is effectively similar to a 256-bit nonce.)

paragonie-scott:
The lastest commit should satisfy #93

paragonie-scott:
Forgot about the runtime tests. Need new test vectors.

doc/02-Crypto.php.md
```diff
@@ -0,0 +1,101 @@
+Symmetric Key Encryption
+========================
+
+At a glance:
+
+* **Cipher and Mode**: `AES-128-CBC`
```
defuse: This is now CTR.

doc/02-Crypto.php.md
```diff
@@ -0,0 +1,101 @@
...
+* **Padding**: `PKCS#7`
```
defuse: Not used since CTR.

paragonie-scott:
One thing that the `File` class doesn't have, but the `Crypto` class does, is the runtime tests. Would you prefer that they be run in that instance as well? (I'd wager your answer is, "Yes".)

doc/02-Crypto.php.md
```diff
@@ -0,0 +1,101 @@
...
+```php
+try {
+    $key = \Defuse\Crypto\Crypto::createNewRandomKey();
+    // WARNING: Do NOT encode $key with bin2hex() or base64_encode(),
+    // they may leak the key to the attacker through side channels.
+    // 
+    // This library offers these two for cache-timing-safe encoding:
+    //
+    //     \Defuse\Crypto\Crypto\binToHex()
+    //     \Defuse\Crypto\Crypto\hexToBin()
```
defuse: Awesome!

doc/02-Crypto.php.md
```diff
@@ -0,0 +1,101 @@
...
+    // Either:
+    //   1. The ciphertext was modified by the attacker,
+    //   2. The key is wrong, or
+    //   3. $ciphertext is not a valid ciphertext or was corrupted.
+    // Assume the worst.
+    die('DANGER! DANGER! The ciphertext has been tampered with!');
+} catch (CryptoTestFailedException $ex) {
+    die('Cannot safely perform decryption');
+} catch (CannotPerformOperationException $ex) {
+    die('Cannot safely perform decryption');
```
defuse: An important property to keep in mind: Tampering the ciphertext should *only* be able to lead to InvalidCiphertext exceptions, i.e. the attacker shouldn't be able to trick you into thinking the runtime test failed or something like that (just thinking out loud).

doc/02-Crypto.php.md
```diff
@@ -0,0 +1,101 @@
...
+}
+```
+
+Decrypting a Message
+--------------------
+
+```php
+try {
+    $decrypted = self::Decrypt($ciphertext, $key);
+} catch (\Defuse\Crypto\Exception\InvalidCiphertext $ex) { // VERY IMPORTANT
```
defuse: This exception seems to be in a different namespace than the ones below, is it still correct?

doc/02-Crypto.php.md
```diff
@@ -0,0 +1,101 @@
...
+try {
+    $decrypted = self::legacyDecrypt($ciphertext, $key);
+} catch (\Defuse\Crypto\Exception\InvalidCiphertext $ex) { // VERY IMPORTANT
+    // Either:
+    //   1. The ciphertext was modified by the attacker,
+    //   2. The key is wrong, or
+    //   3. $ciphertext is not a valid ciphertext or was corrupted.
+    // Assume the worst.
+    die('DANGER! DANGER! The ciphertext has been tampered with!');
+} catch (CryptoTestFailedException $ex) {
```
defuse: I think these might be in a different namespace now?

src/Core.php
```diff
@@ -0,0 +1,394 @@
+<?php
+namespace Defuse\Crypto;
+
+use \Defuse\Crypto\Exception as Ex;
+
+class Core
+{
+    const VERSION = "\xD3\xF5\x02\x00";
```
defuse: What is the meaning of these bytes? What will the next version number be?

src/File.php
```diff
@@ -0,0 +1,742 @@
...
+final class File extends Core implements StreamInterface
+{
+    /**
+     * Use this to generate a random encryption key.
+     *
+     * @return string
+     */
+    public static function createNewRandomKey()
+    {
+        $config = self::getVersionConfig(self::VERSION);
```
defuse: Do file ciphertexts start with the same 'version' as normal ciphertexts? There should be a way to tell them apart using the version number (or something) and an exception thrown if you try to decrypt with the wrong API (e.g. normal ciphertext saved to a file).

src/File.php
```diff
@@ -0,0 +1,742 @@
...
+                    'Cannot open input file for encrypting'
+                );
+            }
+            \stream_set_read_buffer($if, 0);
+
+            /**
+             * Output file handle
+             */
+            $of = \fopen($outputFilename, 'wb');
+            if ($of === false) {
```
defuse: `$if` doesn't get closed here.

src/File.php
```diff
@@ -0,0 +1,742 @@
...
+                    'Cannot open input file for decrypting'
+                );
+            }
+            \stream_set_read_buffer($if, 0);
+
+            /**
+             * Output file handle
+             */
+            $of = \fopen($outputFilename, 'wb');
+            if ($of === false) {
```
defuse: `$if` doesn't get closed here.

src/File.php
```diff
@@ -0,0 +1,742 @@
...
+     *
+     * @param string $header
+     */
+    protected static function getVersionConfig($header)
+    {
+        $valid = 0;
+        $valid |= $header[0] ^ "\xDE";
+        $valid |= $header[1] ^ "\xF5";
+        $major = \ord($header[2]);
+        $minor = \ord($header[3]);
```
defuse: Ah, here's the code that defines what the version means.

defuse:
I haven't looked too closely at the issues, but I'm going to merge this into a v2.0 branch so that it's easier to work with, and then open tickets for the issues I found.

src/File.php
paragonie-scott: That's a good idea. Would you be fine with the last bit of the second byte being used to specify file vs string?

src/Core.php
paragonie-scott: Either `"\xD3\xF5\x02\x01"` or `"\xD3\xF5\x03\x00"` depending on how the library changes over time. :)

doc/02-Crypto.php.md
paragonie-scott: Yes! Good catch.

paragonie-scott:
Okay, sounds good to me. Need me to resubmit the PR?

defuse:
@paragonie-scott: I just created the v2.0 branch, so feel free to submit PRs to it if you feel like fixing any of those things. :)

defuse:
Hang on, I'm still trying to figure out how to push that branch to github.

defuse:
There we go.

paragonie-scott:
This has been merged in another branch.

src/File.php
defuse: I opened #105 for that. Yes. So it would be DEF4 and DEF5?

src/File.php
defuse: Hey, it's "DE" here but "D3" up there! How is it working!?

PR #81, opened 2015-07-23
Add interface and instantiatable class. Closes #79 by @Rican7

### LINKED ISSUE #79: Use of statics prohibits testing/injection/mocking capabilities
This library uses all static methods, and while some of them make sense, its really too bad that the library's Crypto class can't be used as an injectable instance. Currently, testing services that make use of this library means running expensive crypto hashing algorithms that slow down a unit testing feedback loop.

Switching to non-statics would also allow for certain configurations to be used, if certain aspects needed to be overrided (I know, likely not suggested), for example providing a different hash function but defaulting to the current constant `sha256`.

Thank you for this great library and for considering this issue.

paragonie-scott:
See the discussion in #79


src/CryptoInterface.php
```diff
@@ -0,0 +1,9 @@
+<?php
+namespace Defuse\Crypto;
+
+interface CryptoInterface
+{
+    public function createNewRandomKey();
+    public function encrypt($plaintext, $key);
+    public function decrypt($ciphertext, $key);
+}
```
Rican7: We should probably copy the method docs of the `Crypto` class methods onto these methods. Then implementations can just `{@inheritdoc}`. :smiley:

defuse:
I am instituting an outright ban on things like "Closes #79" in commit messages. I am usually too busy looking at the code to realize that merging it will close an issue, and for human-reported issues I like to ask them if the changes solved their problem before closing the issue.

For this one, it's okay, there's no need to change the commit message. But for the future please no magic! Thanks!

paragonie-scott:
Understood!

defuse:
This can optionally be updated for 2.0, but I don't think it's a high priority (we can always add it in 2.1 without affecting the existing API).

paragonie-scott:
I can refactor slightly for 2.0 and resubmit it before we release 2.0. It isn't a high priority, but no reason to eschew it.

src/CryptoInterface.php
```diff
@@ -0,0 +1,40 @@
...
+     * 
+     * Encrypts a message.
+     * $plaintext is the message to encrypt.
+     * $key is the encryption key, a value generated by CreateNewRandomKey().
+     * You MUST catch exceptions thrown by this function. See docs above.
+     * 
+     * @param string $plaintext
+     * @param string $key
+     * @return string
+     * @throws Ex\CannotPerformOperationException
```
sstok: Missing a `use` for exception alias, `Ex` doesn't exist as sub namespace.

paragonie-scott: Thanks, I'll be sure to fix this when I update/resubmit the PR. :)

defuse:
Is this already in 2.0*? If it is I'm going to close it.

* - This is part of my psychological friction with 2.0, I don't even *know* what's in there. Definitely a lesson learned next time I want to accept outside contributions to a library -- I need to do a better job of staying on top of it.

defuse:
Oh there's a 2.1 label so presumably not :)

defuse:
Actually, I'm rejecting this. There should *not* be two different ways to use the library. Users will wonder which one they should be using, etc. It's unnecessary complexity and we shouldn't have to maintain it. If someone needs this for testing or whatever they can take on the burden of doing it themselves.

defuse:
I'm rejecting this change, especially since I don't have time to spend on this library and want it to go into maintenance mode, for the reasons described here https://github.com/defuse/php-encryption/pull/81#issuecomment-204827652

defuse:
Whoops, meant to post that on a different ticket.

Rican7:
Wow. Ok. Good to know.

defuse:
@Rican7: I'm sorry if this feels like the wrong decision. I should explain more: I'm heading off to grad school in 5 months, and while I'm there I won't have any time to maintain this library beyond bug/security fixes. So I really want to get 2.0 out the door, and avoid having to do a 2.1.

If a *lot* of people really want this then I'll of course revisit this decision. (Even if I don't revisit this decision, the library is under an open license, so anyone's welcome to fork it and add that functionality.)

PR #82, opened 2015-07-23
CannotPerformOperation should be called CannotPerformOperationException

harryjackson:


sstok:
:+1:

piqus:
:+1:  +1000

defuse:
Whoa, sorry about that! What is the point of moving that assignment?

defuse:
Also, it needs to be changed in `example.php` and `tests/runtime.php` as well.

defuse:
I fixed all three places in: d2c33279e31d5f3d4b74a82139b6ce56d551965b

defuse:
I like the place that assignment is, and don't see a reason to move it, so I'm closing this PR (feel free to re-open it if you think that's a mistake).

harryjackson:
The reason I moved the assignment is because the autoloader for php-encryption is require to run before the Autoloader in our Yii App or it won't work, this means thousands of calls to this autoloader, per request,  has this redundant assignment ie the "if" statement almost always evaluates to true making the assignment redundant. I didn't see an aesthetic or logical reason to keep it where it was so I moved it. It's up to you what you do but I would typically not do the following...

$var = "some variable";
if($foo != "bar") {
  return; //Note, we have not used $var because it's not required in this condition
}
...
$blah = blah($var);

I would almost always do this...

if($foo != "bar") {
  return; //Note, we have not used $var because it's not required in this condition
}
$var = "some variable"; //Assuming this assignment adds clarity we define this variable here
...
$blah = blah($var); //Then use it

PR #87, opened 2015-08-06
Create LICENSE file

paragonie-scott:
Makes it clearer to people what the terms of its release are, even though this is in composer.json

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,25 @@
+PHP Encryption Library
+Copyright (c) 2014-2015, Taylor Hornby <https://defuse.ca>
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are met:
+
+1. Redistributions of source code must retain the above copyright notice,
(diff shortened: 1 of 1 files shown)


PeeHaa:
Sorry to hijack this PR, but is there a specific reason the "All rights reserved" line is included?

adamcaudill:
@PeeHaa That's a standard part of the BSD license.

PeeHaa:
Thanks for clarifying. The reason I didn't expect that line is because I was under the assumption that the library was [MIT licensed instead](https://github.com/defuse/php-encryption/blob/master/composer.json#L4).

Could you please fix the composer.json file or do you want me to open a separate issue for it?

Either way here is a token of my appreciation in the form of a [friday link](https://www.youtube.com/watch?v=kfVsfOSbJY0) for your excellent work on this library.

paragonie-scott:
I'm going to close this PR because I'm not confident that my change is correct.

@defuse If you can create a LICENSE file (and update the composer.json, if necessary), that would probably clarify the licensing for everyone. :)

PR #92, opened 2015-09-28
modified condition to avoid an inspection error in PHPStorm

karlpatrickespiritu:
screenshot from 2015-09-28 23 47 37

### DIFF
--- a/src/ExceptionHandler.php
+++ b/src/ExceptionHandler.php
@@ -39,7 +39,7 @@ class ExceptionHandler
     }
 
     public function __destruct() {
-        if ($this->rethrow) {
+        if ($this->rethrow instanceof \Exception) {
             throw $this->rethrow;
         }
     }
--- a/src/ExceptionHandler.php
+++ b/src/ExceptionHandler.php
@@ -17,6 +17,7 @@ use \Defuse\Crypto\Exception as Ex;
 
 class ExceptionHandler
 {
+    /* @var \Exception */
     private $rethrow = NULL;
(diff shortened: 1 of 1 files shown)


paragonie-scott:
This reminds me of leaving a bunch of `if (false) { $object = new Explicit\Object\Path\Name; }` in my PHP scripts to make Netbeans ctrl+click through to the desired method.

That said, :+1: we don't need to worry about `\Throwable` or `\Error`, because even when PHP 7 is out, all of the exceptions here extend `\Exception`.

sethbattin:
There's an alternate way to make phpstorm recognize the object, without a code change.  Just add an annotation comment to the private variable:

`/* @var \Exception */
private $rethrow */

If code changes are undesirable, then make the code inspector leverage it's additional advantage.

paragonie-scott:
> If code changes are undesirable, then make the code inspector leverage it's additional advantage.

That's a good point too. Given these two solutions, the annotation should accomplish the same result with less runtime overhead. (Note to self: Take advantage of annotations.)

karlpatrickespiritu:
the @var \Exception works like a charm. didn't know about that. thanks.

karlpatrickespiritu:
didn't know about the /* @var \Exception */ just now. thanks for the info. and good point, less runtime overhead :+1:

sstok:
:+1:

defuse:
We're going to get rid of that exception handler override thing (#97), so I'm closing this.

PR #116, opened 2015-10-16
v2.0.remove-unnecessary-inheritance.0

defuse:
See the commit messages.


defuse:
@paragonie-scott: It would be *wonderful* if you could review this PR, and, at your leisure, argue against removing the inheritance if you think it's valuable.

defuse:
(What it does is written in the commit message)

paragonie-scott:
Aside from the failure caused by #22 (and possibly any errors that failure is hiding), this looks very good to me. I like the way you organize everything. :+1:

paragonie-scott:
>  It would be *wonderful* if you could review this PR, and, at your leisure, argue against removing the inheritance if you think it's valuable.

The only reason I used inheritance in my initial PR #78 was to expose common methods that the derived classes (`Crypto` and `File) could use, but were not public. It's not a feature I feel super strongly about, and someone could always make

    class CryptoProxy extends \Defuse\Crypto\Core {
        public function getInaccessibleMethod($method, $args = []) {
            return parent::$method(...$args);
        }
    }

and there's nothing inheritance can do to stop this.

src/File.php
```diff
@@ -393,28 +416,31 @@ public static function decryptResource($inputHandle, $outputHandle, $key)
         $remaining = 4;
```
paragonie-scott: For this, shouldn't we use the header length constant?

src/Key.php
```diff
@@ -0,0 +1,188 @@
...
+        $this->config = self::GetKeyVersionConfigFromKeyHeader($this->key_version_header);
+    }
+
+    public function saveToAsciiSafeString()
+    {
+        return Encoding::binToHex(
+            $this->key_version_header .
+            $this->key_bytes .
+            hash(
+                $this->config['CHECKSUM_HASH_FUNCTION'],
```
paragonie-scott: Should this also be converted from an array to an object?

paragonie-scott:
Aside from a few very minor things this looks good to me. I do have one suggestion but I'll open an issue for that.

defuse:
Okay, fixed those minor issues, and the travis testing is passing. I think it's time to land this!

PR #128, opened 2015-10-19
Minor Boyscouting for v2.0

paragonie-scott:
Remove unused "use" statements
Add docblocks
Fix a spelling error

### DIFF
--- a/src/RuntimeTests.php
+++ b/src/RuntimeTests.php
@@ -12,7 +12,7 @@ use \Defuse\Crypto\Crypto;
  * class. Always prefix explicitly with Crypto:: or RuntimeTests::.
  */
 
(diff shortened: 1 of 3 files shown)


src/RuntimeTests.php
```diff
@@ -219,12 +230,14 @@ private static function AESTestVector($config)
 
         $computed_ciphertext = Crypto::plainEncrypt($plaintext, $key, $iv, $config);
         if ($computed_ciphertext !== $ciphertext) {
+            /*
```
paragonie-scott: I can revert this if you still need it for testing.

defuse: We can actually delete all those commented-out lines.

Issue #125, opened 2015-10-18
Prevent Partial Reads and Writes


paragonie-scott:
See also: [Halite's `File::readBytes()` and `File::writeBytes()`](https://github.com/paragonie/halite/blob/ee5e92ee65dfebf26b7d8c559921ce2a5d4cf76c/src/File.php#L973-L1041) API for reference.

This could greatly reduce the amount of `if ($result === false)` lines by making a separate method for stream reading/writing that throws an exception if it can't proceed.

defuse:
That looks good. I'll take it and modify it to fit within this project. Thanks!

defuse:
Oh there might already be a PR for this issue.

paragonie-scott:
Yeah, I made #129 a few hours ago. :P

paragonie-scott:
This is a duplicate of #114

Issue #114, opened 2015-10-16
In File, length of returned values are not checked.


defuse:
```
            $ivsize = \openssl_cipher_iv_length($config['CIPHER_METHOD']);
            $iv = \fread($inputHandle, $ivsize);
            if ($iv === false ) {
                throw new Ex\CannotPerformOperation(
                    'Cannot read input file'
                );
            }
```

None of the code below that checks if `$iv` is actually the correct length, so if `EOF` was encountered (or some other weird condition), we'll get a wrong size IV. The same bug probably exists for other parameters read from the file too.

paragonie-scott:
Feel free to copy https://github.com/paragonie/halite/blob/b202f427062c29566b6ae09b1ee72b2bc1746d61/src/File.php#L667-L695 for file reads.

paragonie-scott:
(Aside: I'm going to create a congruent method in Halite for file writes.)

paragonie-scott:
Should be fixed in #129

defuse:
This looks good, except for the issue in #145.

PR #129, opened 2015-10-19
Prevent partial reads/writes

paragonie-scott:
Reference: #125

### DIFF
--- a/src/File.php
+++ b/src/File.php
@@ -290,15 +290,11 @@ final class File implements StreamInterface
         /**
          * First let's write our header, file salt, and IV to the first N blocks of the output file
          */
-        if (\fwrite(
+        self::writeBytes(
             $outputHandle,
             Core::CURRENT_FILE_VERSION . $file_salt . $iv, 
             Core::HEADER_VERSION_SIZE + $config->saltByteSize() + $ivsize
-        ) === false) {
-            throw new Ex\CannotPerformOperationException(
-                'Cannot write to output file'
-            );
-        }
+        );
 
         /**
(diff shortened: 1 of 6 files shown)


src/File.php
```diff
@@ -799,4 +773,85 @@ private static function getFileVersionConfigFromMajorMinor($major, $minor)
...
+        $buf = '';
+        $remaining = $num;
+        do {
+            if ($remaining <= 0) {
+                break;
+            }
+            $read = \fread($stream, $remaining);
+            if ($read === false) {
+                throw new Ex\CannotPerformOperationException(
+                    'Could not write to the file'
```
paragonie-scott: Oops. Need to fix the text here.

paragonie-scott:
Issues that should be fixed now:

* #114 
* #132 
* #133 
* Possibly #134 (thought we might want to keep that open until audited)

defuse:
Fuck, I want to reply to one of @paragonie-scott's comments, but I can't remember which commit it's in, and GitHub doesn't give me the link.

> No, it doesn't return false. You threw a return; in encryptFile() and the unit tests didn't catch it.

So I accidentally added a return statement somewhere in the function, and that broke it, without breaking the unit tests, and returning true ensures that they'll catch it in the future? Could you document that (or just point me to a commit), I'd like to improve the unit tests so that they'll catch it.

defuse:
@paragonie-scott: I think you addressed all my comments, is this ready to merge?

paragonie-scott:
I believe it's ready.

PR #96, opened 2015-10-16
Harden the autoloader against unserialize -> LFI risks.

paragonie-scott:

PR #140, opened 2015-12-10
Harden the autoloader against unserialize -> LFI risks

paragonie-scott:
Replaces #96 (also updated classmap)

### DIFF
--- a/autoload.php
+++ b/autoload.php
@@ -4,7 +4,7 @@
  */
 \spl_autoload_register(function ($class) {
     // Project-specific namespace prefix
-    $prefix = 'Defuse\\Crypto';
+    $prefix = 'Defuse\\Crypto\\';
 
     // Base directory for the namespace prefix
     $base_dir = __DIR__.'/src/';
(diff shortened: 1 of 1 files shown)


autoload.php
```diff
@@ -18,19 +18,46 @@
...
+            'Exception/CannotPerformOperationException.php',
+        'Exception\\CryptoException' =>
+            'Exception/CryptoException.php',
+        'Exception\\CryptoTestFailedException' =>
+            'Exception/CryptoTestFailedException.php',
+        'Exception\\InvalidCiphertextException' =>
+            'Exception/InvalidCiphertextException.php',
+    );
+    foreach ($classmap as $classname => $file) {
+        if ($classname === $relative_class) {
```
StanAngeloff: `isset($classmap[$relative_class])` seems to do the work just as well?

defuse: Yeah, it really doesn't need to be a loop. But I'm going to accept it and open a ticket to fix it later.

defuse:
LGTM

PR #143, opened 2015-12-11
restore exception handler when cleaning up object

Lewiscowles1986:


defuse:
Is there any reason to do this? If this class has been used properly, and object is getting destructed without throwing an exception it (should) mean the execution is terminating, and there's no reason to restore the handler.

defuse:
I guess you could have a scope where you're using the library and you want to instantiate this class, use the library, then let the object go out of scope.

defuse:
(This whole stupid mechanism gets removed in 2.0 anyway).

Lewiscowles1986:
Thx,

You didn't have to merge, I'd forgotten about it, but yeah someone complained about another lib using this in such a way it did not terminate; so I fixed the mechanism, thought I'd push to your lib. It's been a while and I don't even remember which project it was on; apologies for not providing a good commit message etc.

Issue #144, opened 2015-12-12
Convert File tests to PHPUnit


defuse:
The file tests are currently individual PHP files inside `test/stream`. It would be nice if they were all converted to PHPUnit tests. That could be difficult, but I think worth doing.

defuse:
Make sure to test `File::createNewRandomKey()`. I don't think anything's exersizing that currently.

defuse:
This is done.

PR #147, opened 2015-12-12
Issue/144 convert file tests

sethbattin:
I saw issue #144 fly by my inbox and decided to try to do it.  Let me know if I've missed anything glaring; I've never looked closely at the library before.

Note that my last test method, `testFileCreateRandomKey`, is failing.  I strongly suspect that the method being tested (`File::createNewRandomKey`) is a relic that should be removed.  It has a name that is identical to other code, but it doesn't function the same way.  If I'm right, I could add another commit to remove those parts.


defuse:
This is awesome. They're even catching new bugs! I'll review this right away :D

test/unit/FileTest.php
```diff
@@ -0,0 +1,154 @@
...
+    {
+        $data = [];
+        
+        $data['wat-giagantic-duck'] = ['wat-gigantic-duck.jpg'];
+        $data['large'] = ['large.jpg'];
+        
+        if (file_exists(__DIR__ . '/File/In_the_Conservatory.jpg')){
+            // see File/get_large.sh
+            $data['extra-large'] = ['In_the_Conservatory.jpg'];
+        }
```
defuse: I wonder why that file isn't checked in. I'm not even sure where to get it.

paragonie-scott: get_large.sh

defuse: Oh I'm an idiot... there's a comment right there telling me to look in get_large.sh.

defuse:
Looks good!

test/unit/FileTest.php
paragonie-scott: It's too large for github. The file tests were originally intended for
benchmarking memory usage BTW. I welcome making these unit tests :)

PR #152, opened 2015-12-15
Add helper method for openssl_cipher_iv_length().

twistor:


paragonie-scott:
> 1) Defuse\Crypto\FileTest::testFileCreateRandomKey
> 
> Failed asserting that Binary String: 0xec5a666a1d861...c6fcc64 is an instance of class "\Defuse\Crypto\Key".

I'm not sure where this change-set caused the reported failure (was it at the same time as another change?). This is a good idea, however.

twistor:
That's not from this patch.

I get the same failure running v2.0 locally.

paragonie-scott:
Okay, well this looks great to me. :+1:

defuse:
@paragonie-scott Yeah, that was a bug. I think it's fixed in the v2.0 branch now.

PR #153, opened 2015-12-15
Test for qualified functions

twistor:
This is a quick proof of concept I hacked together.

I would guess the changes to 'test/phpunit.sh' are unwanted. We should be able to manually include the autoloader from the test class.

Also, the test directory is excluded for time being. That should probably be fixed as well.

Posting here to get feedback before moving on.


defuse:
Closing this because the stupid github bug is preventing me from being able to review this. I will re-submit the same PR myself.

defuse:
That didn't work. I reviewed it by [looking at this](https://github.com/defuse/php-encryption/compare/v2.0...twistor:test-for-qualified-functions?expand=1).

defuse:
Resubmitted as #186.

PR #157, opened 2015-12-27
Make autoload function conform to spec.

twistor:
According to http://php.net/manual/en/function.spl-autoload-register.php, the autoload function should return true on success, and false on failure.

Also, changed the loop to a simple isset(). If there was a reason for the loop, I can change it back.


paragonie-scott:
:+1:

sethbattin:
I think this change is misguided.  The linked documentation refers to true/false return values when registering the function, i.e. `apl_autoload_register`'s return value.  The function being registered should not return any value, because the existence of the loaded class defines success or failure (and whether to continue autoloading).

[PSR4 says that an autoload function should not return a value whatsoever.](http://www.php-fig.org/psr/psr-4/)

glensc:
autoload function doing return is waste of cycles. the php-core implementation checks if function returned value, and does extra code to free that result never using it.

https://github.com/php/php-src/blob/php-7.0.2/ext/spl/php_spl.c#L427-L430

twistor:
Yup. It ended up being a bug in composer that dragged me down this rabbit hole.

I haven't closed this yet because I was planning on still proposing removing the foreach.

defuse:
Closing this PR because it appears to be broken.

Issue #161, opened 2016-01-30
key based on password


robstoll:
Hi, 

I have the following scenario: 
- data is saved encrypted in DB
- key for encryption must not be saved directly on the system (in clear text)
- yet, application needs the key to decrypt the data
- hence key shall be saved encrypted based on the users password

therefore I would like to know, how can I generate a key passed on a users password?

Cheers,
Robert

sarciszewski:
[`hash_pbkdf2()`](https://secure.php.net/hash_pbkdf2).

robstoll:
and how can I use the generated key with hash_pdkdf2 with your Crypto::encrypt()?

sarciszewski:
```php
$key = hash_pbkdf2(
    'sha256',
    $YOUR_USERS_PASSWORD_HERE,
    $A_HARD_CODED_PASSWORD_MADE_OF_32_RANDOM_BYTES_UNIQUE_PER_USER,
    100000, // minimum 86000
    16,
    true
);

$cipher = Crypto::encrypt($message, $key);
$plain = Crypto::decrypt($cipher, $key);

var_dump($message === $cipher); // bool(true)
```

sarciszewski:
Also, you want to use a CSPRNG for the salt. `Crypto::createRandomKey()` returns 16 bytes. That should be acceptable. Store that long-term per user.

robstoll:
Thanks for the quick reply. I should have mentioned that I am using version 2.x of your library. Crypto::encrypt incorporates the following check:
{code}
if (!is_a($key, "\Defuse\Crypto\Key")) {
            throw new Ex\CannotPerformOperationException(
                "The given key is not a valid Key object."
            );
        }
{code}
Hence I need to transform the key generated by hash_pbkdf2 into a \Defuse\Crypto\Key somehow. 
Any idea?

sstok:
Use `Key::saveToAsciiSafeString` to convert to the Key object to a string for storing, then use `Key::LoadFromAsciiSafeString` to convert the stored string back to a Key object.

robstoll:
@sstok I saw this functionality but it does not help me since I generate the key with hash_pbkdf2 which returns a string (or raw binary data).

robstoll:
For now I am using the method I added in pull request #163 
Please let me know if there is a better way to do this without breaking backward compatibility

pnowosie:
What you need is a long planed feature #5, not implemented yet in v2.0. Maybe your's PR will push things forward a little.

[...]

PR #163, opened 2016-01-31
create key based on password

robstoll:
added the method CreateKeyBasedOnPassword which uses hash_pbkdf2 to generate a key.
For a use case, see #161


defuse:
GitHub is broken so I'm going to try to resumbit this pull request.

Issue #165, opened 2016-02-06
encrypt and decrypt empty string


robstoll:
I am getting the following error
`Defuse\Crypto\Exception\InvalidCiphertextException with message 'Ciphertext is too short.'`
When I do the following:

```php
use Defuse\Crypto\Crypto;
$key = Crypto::CreateNewRandomKey();
$crypt= Crypto::encrypt('', $key, true);
$plain = Crypto::decrypt($crypt, $key, true);
```
Is this a bug? A security precaution? If so, why is it possible to encrypt an empty string in the first place?

paragonie-scott:
> There are two hard things in computer science: cache invalidation, naming things, and off-by-one errors.

Pull request #166 contains a unit test and patch, should we consider "able to encrypt/decrypt empty strings" to be expected behavior.

robstoll:
I am no security expert and hence do not know if information is leaked out while encrypting and decrypting empty strings. If that is not the case, then I do not really see why this should not be expected behaviour

paragonie-scott:
https://3v4l.org/uULd5

robstoll:
Looks good to me, but I am no security expert. Is there a reason why the library does not support it in the first place?

paragonie-scott:
> Is there a reason why the library does not support it in the first place?

That code originally worked on a block cipher in CBC mode, which means the ciphertext size would always be at least 1 block long (16 bytes). Version 2 (which, by the way, you shouldn't be using yet) is switching to CTR mode, which doesn't impose this requirement.

robstoll:
> which, by the way, you shouldn't be using yet
oh really, is 2.x not stable yet? How easy will it be to migrate from 1.x to 2.x? I am using your library in a new project and do not want to migrate to 2.0 after a year or so

paragonie-scott:
> oh really, is 2.x not stable yet?

Nope. It's still a [work in progress](https://github.com/defuse/php-encryption/releases).

> How easy will it be to migrate from 1.x to 2.x?

We have a [`legacyDecrypt()`](https://github.com/defuse/php-encryption/blob/v2.0/src/Crypto.php#L237-L321) method to decrypt old messages so you can re-encrypt them with the new ciphertext format. But one of the big differences is that, instead of a binary string, the encrypt methods now expect a [`Key`](https://github.com/defuse/php-encryption/blob/v2.0/src/Key.php) object.

It will require a little bit of effort, but not a huge reverse-engineering effort to access your old data.

[...]

PR #166, opened 2016-02-07
Decrypt empty strings, with unit test.

paragonie-scott:
Reference #165

### DIFF
--- a/src/Crypto.php
+++ b/src/Crypto.php
@@ -217,7 +217,7 @@ class Crypto
                     "Could not get the IV length from OpenSSL"
                 );
             }
-            if (Core::ourStrlen($ciphertext) <= $ivsize) {
+            if (Core::ourStrlen($ciphertext) < $ivsize) {
                 throw new Ex\InvalidCiphertextException(
                     "Ciphertext is too short."
                 );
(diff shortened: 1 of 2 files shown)


test/unit/CryptoTest.php
```diff
@@ -0,0 +1,17 @@
...
+use \Defuse\Crypto\Key;
+
+class CryptoTest extends PHPUnit_Framework_TestCase
+{
+    public function testEmptyString()
+    {
+        $str = '';
+        $key = Key::CreateNewRandomKey();
+        $cipher = Crypto::encrypt($str, $key);
+        $this->assertEquals(
```
sstok: Hehe it's actually $expected, $result :) I two make this mistake all the time.
https://phpunit.de/manual/current/en/appendixes.assertions.html#appendixes.assertions.assertEquals

defuse:
Why was this closed?

src/Crypto.php
```diff
@@ -226,6 +226,14 @@ public static function decrypt($ciphertext, $key, $raw_binary = false)
...
                 throw new Ex\CannotPerformOperationException();
             }
+            /**
+             * If you encrypt an empty string with a stream cipher, you get an
+             * empty string. So if we have exactly $ivsize bytes, we know that
+             * is what happened here.
+             */
+            if (Core::ourStrlen($ciphertext) === $ivsize) {
+                return '';
+            }
```
defuse: @paragonie-scott: You might like to know for future reference that this would introduce a security bug. This is before the HMAC check, so this allows an attacker to transform any valid ciphertext into a valid ciphertext of the empty string by truncating it. The right thing to do here is to just let the code below run through (decrypting the empty string) instead of trying to handle it as a separate case.

PR #167, opened 2016-02-07
Decrypt empty strings v2.0

paragonie-scott:


paragonie-scott:
God dammit git. I'm just going to start over.

PR #169, opened 2016-02-12
Something for the version 1.x branch

paragonie-scott:
* Update README to include hyperlinks
* Throw an exception if you pass a non-string to Crypto::encrypt() or Crypto::decrypt()


paragonie-scott:
(This is totally optional, by the way.)

PR #170, opened 2016-02-12
Fix unit tests

paragonie-scott:


src/File.php
```diff
@@ -771,10 +785,12 @@ private static function getFileVersionConfigFromMajorMinor($major, $minor)
      */
     final public static function readBytes($stream, $num)
     {
-        if ($num <= 0) {
+        if ($num < 0) {
```
defuse: Good catch! It's definitely valid to try and read 0 bytes.

src/File.php
```diff
@@ -400,6 +408,12 @@ public static function decryptResource($inputHandle, $outputHandle, Key $key)
                 'Output handle must be a resource!'
             );
         }
+        $stat = \fstat($inputHandle);
```
defuse: I guess we can't (or it's unreasonably complex) do this in a TOCTTOU-free way.

src/Exception/InvalidInput.php
```diff
@@ -0,0 +1,7 @@
+<?php
+namespace Defuse\Crypto\Exception;
+
+class InvalidInput extends \Defuse\Crypto\Exception\CryptoException
```
defuse: This PR adds this exception, but where does it get used?

defuse:
Will merge once the `InvalidInput` question is answered :)

src/Exception/InvalidInput.php
paragonie-scott: https://github.com/paragonie/php-encryption/blob/fix-taylors-dec12-branch/src/File.php#L407

It looks like there was already code written for these exceptions without defining it.

defuse:
Thank you!!

PR #172, opened 2016-02-13
Separate v2

paragonie-scott:
Fix the v2.0 branch


paragonie-scott:
Did I seriously just do that? Merge to master?

PR #180, opened 2016-03-12
Version 1.x - Cache-timing safety

paragonie-scott:
https://github.com/phpseclib/phpseclib/issues/761

### DIFF
--- a/src/Crypto.php
+++ b/src/Crypto.php
@@ -700,10 +700,14 @@ final class Crypto
         $hex = '';
         $len = self::ourStrlen($bin_string);
         for ($i = 0; $i < $len; ++$i) {
-            $c = \ord($bin_string[$i]) & 0xf;
-            $b = \ord($bin_string[$i]) >> 4;
-            $hex .= \chr(87 + $b + ((($b - 10) >> 8) & ~38));
-            $hex .= \chr(87 + $c + ((($c - 10) >> 8) & ~38));
+            $chunk = unpack('C', self::ourSubstr($bin_string, $i, 1));
+            $c = $chunk[1] & 0xf;
+            $b = $chunk[1] >> 4;
+            $hex .= pack(
+                'CC',
+                (87 + $b + ((($b - 10) >> 8) & ~38)),
+                (87 + $c + ((($c - 10) >> 8) & ~38))
+            );
         }
         return $hex;
     }
@@ -722,8 +726,10 @@ final class Crypto
         $hex_len = self::ourStrlen($hex_string);
         $state = 0;
         
+        $chunk = \unpack('C*', $hex_string);
         while ($hex_pos < $hex_len) {
-            $c = \ord($hex_string[$hex_pos]);
+            ++$hex_pos;
+            $c = $chunk[$hex_pos];
             $c_num = $c ^ 48;
             $c_num0 = ($c_num - 10) >> 8;
             $c_alpha = ($c & ~32) - 55;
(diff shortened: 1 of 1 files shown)


paragonie-scott:
I'm also going to send the same PR against the v2 branch, so if this isn't a priority, feel free to close/reject.

src/Crypto.php
```diff
@@ -722,8 +726,10 @@ public static function hexToBin($hex_string)
         $hex_len = self::ourStrlen($hex_string);
         $state = 0;
         
+        $chunk = \unpack('C*', $hex_string);
         while ($hex_pos < $hex_len) {
-            $c = \ord($hex_string[$hex_pos]);
+            ++$hex_pos;
+            $c = $chunk[$hex_pos];
```
defuse: This accesses index 1 first. Shouldn't it be index 0 like the old code?

defuse: Nope, it should be 1:

```
[firexware@firexware-pc ~]$ php -r "var_dump(unpack('C*', '414243'));"
Command line code:1:
array(6) {
  [1] =>
  int(52)
  [2] =>
  int(49)
  [3] =>
  int(52)
  [4] =>
  int(50)
  [5] =>
  int(52)
  [6] =>
  int(51)
}
```

Why on earth...? Whatever.

defuse:
`ord()` appears to be timing safe:

https://github.com/php/php-src/blob/c72282a13b12b7e572469eba7a7ce593d900a8a2/ext/standard/string.c#L2767

`chr()` is not, at least if this branch is taken:

https://github.com/php/php-src/blob/c72282a13b12b7e572469eba7a7ce593d900a8a2/ext/standard/string.c#L2794

defuse:
I don't buy the reasoning in https://github.com/phpseclib/phpseclib/issues/761#issuecomment-127269694. The code presented just shows that with `ord()` you have to write the loop in PHP and with `unpack()` the loop happens inside the call to `unpack()`. It says nothing about what `ord()` and `unpack()` do, and whether what they do leaks information about the input string.

Note that we *don't* care about the lookups into the string for `ord()`, since the indexes are not secret (it's just a counter). What we *do* care about is whether `ord()` leaks the value of the character it was passed, and a quick look at the [relevant PHP source code](https://github.com/php/php-src/blob/c72282a13b12b7e572469eba7a7ce593d900a8a2/ext/standard/string.c#L2767) suggests it doesn't, it just casts the a character to a long.

`chr()` does use a table lookup which leaks information about the character if it takes [this branch](https://github.com/php/php-src/blob/c72282a13b12b7e572469eba7a7ce593d900a8a2/ext/standard/string.c#L2794). I think what it's doing there is that it has pre-defined objects for all possible 1-character strings, and it's optimized to go get one of those instead of creating a new one.

What about `unpack()`? There's a [table lookup](https://github.com/php/php-src/blob/c47c526290b924a1588f317abce94ca3eeddb980/ext/standard/pack.c#L538) but is the index something we don't want to be leaked? We don't care if `i` is leaked since it's just a counter and the attacker already knows it. What about the `map` array? Looking around [here](https://github.com/php/php-src/blob/c47c526290b924a1588f317abce94ca3eeddb980/ext/standard/pack.c#L839) we see that the map parameter is dependent on the type of unpacking we're doing (e.g. "C" vs "V"), but again we don't care if that gets leaked. `unpack()` is not timing safe for things like "H" but at least for "C" it appears to be safe.

Ok, last is `pack()`. It also [looks safe](https://github.com/php/php-src/blob/c47c526290b924a1588f317abce94ca3eeddb980/ext/standard/pack.c#L93) at least for "C".

So, yes, this PR is an improvement over using `ord()` and `chr()`.

paragonie-scott:
:+1:

PR #185, opened 2016-03-30
Fix rethrow class member type.

acbramley:
Without this, IDEs (such as PHPStorm) display an error:

`The thrown object must be an instance of the Exception or Throwable.`


defuse:
I question the sanity of those IDE's, but okay, thanks!

Issue #5, opened 2014-02-06
Derive Key From Password


defuse:
Add a function for deriving a key from a password with PBKDF2.

We have to think *really* hard about how to make this usable, since we want to make sure they're doing the salt right (by making it hard to do it wrong).

sarciszewski:
@htho `hash_pbkdf2` and `hash_equals` can satisfy this need. Luckily, defuse has previously written an implementation for a PBKDF2 class in PHP, so he knows what to do.

defuse:
`password_hash` and `password_verify` are not what you want to make keys with. You need a KDF on its own, like `hash_pbkdf2` (and make sure it's salted, etc).

htho:
Uh I see, my post was offtopic. I was thinking about password hashing for authentication needs, not for encryption.

Anyway, @defuse Have you had a look at `password_hash` and `password_verify`? Are they (in your opinion) convenient for authentication?

pnowosie:
@htho `password_hash` uses `bcrypt` algorithm as such follows 
 [OWASP](https://www.owasp.org/index.php/Password_Storage_Cheat_Sheet) recommendation for authentication / password storage purposes. For `password_verify` I missed the information in docs that it compares hashes in length-constant time witch such function is expected to do.

defuse:
@htho Yes, `password_hash` and `password_verify` are exactly what you should be using for password verification.

defuse:
This is blocking adoption: https://github.com/Webiny/Framework/issues/39#issuecomment-80982903

SvenAlHamad:
From what I've read, the `KeyLocker` should do the trick. It will allow the users to pass a password of any length, and then that can be used to generate a key for the `encrypt` method.

sarciszewski:
@defuse Do you want to assign this one to me? :)

defuse:
Definitely want derive-key-from-password in v2.0.

This should be independently versioned, so that we can start with PBKDF2 and upgrade to Argon2 (or scrypt) later without the user having to special case that themselves.

paragonie-scott:
Feel free to copypasta this: https://github.com/sarciszewski/php-future/blob/be8ef2346341c90e50066db2e6e7be1ea469889d/src/Security.php#L53-L90

I took your old PBKDF2 code and made it PSR-2 compliant (mostly explicitly adding braces after `if` statements).

defuse:
[...]

PR #187, opened 2016-04-02
Password something or other (RESUBMIT)

defuse:


defuse:
This isn't the usable API I imagined in https://github.com/defuse/php-encryption/issues/5#issuecomment-149263138 since it places the burden on the user to securely generate a random salt, instead of it being abstracted away. But whatever. I just want to get v2.0 done.

defuse:
Also, nice improvements, too!

PR #154, opened 2015-12-21
Add more instructions to the README

paragonie-scott:
(Not finished yet.)

PR #190, opened 2016-04-03
Add more instructions to the README (Round Two)

paragonie-scott:
#154 redux

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,6 +13,144 @@ HMAC-SHA256 (Encrypt-then-Mac). HKDF is used to split the user-provided key into
 two keys: one for encryption, and the other for authentication. It is
 implemented using the `openssl_` and `hash_hmac` functions.
 
+## Installing this Library
+
+### Using Composer
+
+```sh
+composer require defuse/php-encryption
+```
+
+### Direct Installation (Phar)
+
+Download the PHP Archive and public key. Extract 
+
+### Direct Installation (Manual)
+
+Download the [latest release](https://github.com/defuse/php-encryption/releases). Extract all of the files into a directory on your webserver (e.g. `/var/www/lib/defuse/php-encryption`).
+
+Then add this to your PHP scripts:
+
+```php
+require '/var/www/lib/defuse/php-encryption/autoload.php';
+```
+
+## Using this Library
+
+1. Generate and store an encryption key.
+2. Encrypt plaintext strings with your key to obtain ciphertext, using `Crypto`.
+3. Decrypt ciphertext strings with your key to obtain plaintext, using `Crypto`.
+4. Encrypt/decrypt files with your key, using `File`.
+
+### Generate and Store an Encryption Key
+
+Generate a new key:
+
+```php
+$key = \Defuse\Crypto\Key::createNewRandomKey();
+````
+
+The above command will generate a random encryption key, using a 
+cryptographically secure pseudorandom number generator. This will generally only
+need to be done *once* if you need to reuse this key for multiple messages.
+
+```php
+$encryptionKeyDataForStorage = $key->saveToAsciiSafeString()
+```
+
+This returns an encoded string that you can use to persist a key across multiple
+runs of the application. You might decide to copy it to a configuration file
+not tracked by Git, for example. To load it again on the next script execution,
+just do this:
+
+```php
+$key = \Defuse\Crypto\Key::LoadFromAsciiSafeString($storedKeyData);
+```
+
+### Encrypting Strings
+
+Once you have a `Key` object, you're ready to encrypt data. All you have to do
+is pass your desired string and the `Key` object to `Crypto::encrypt()`.
+
+```php
+try {
+    $ciphertext = \Defuse\Crypto\Crypto::encrypt("Test message", $key);
+} catch (\Defuse\Crypto\Exception\CryptoTestFailedException $ex) {
+    die("Our platform is not secure enough to use this cryptography library.");
+}
+```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -13,6 +13,144 @@ HMAC-SHA256 (Encrypt-then-Mac). HKDF is used to split the user-provided key into
...
+
+### Using Composer
+
+```sh
+composer require defuse/php-encryption
+```
+
+### Direct Installation (Phar)
+
+Download the PHP Archive and public key. Extract
```
defuse: I think this sentence was supposed to be finished but wasn't?

defuse:
This is really great work!

README.md
paragonie-scott: Yeah. @_@

paragonie-scott:
Thanks. Fixed the snafu. :)

README.md
```diff
@@ -13,6 +13,159 @@ HMAC-SHA256 (Encrypt-then-Mac). HKDF is used to split the user-provided key into
...
+
+### Encrypting Strings
+
+Once you have a `Key` object, you're ready to encrypt data. All you have to do
+is pass your desired string and the `Key` object to `Crypto::encrypt()`.
+
+```php
+try {
+    $ciphertext = \Defuse\Crypto\Crypto::encrypt("Test message", $key);
+} catch (\Defuse\Crypto\Exception\CryptoTestFailedException $ex) {
```
defuse: This could throw a bunch of different types of exceptions, like CannotPerformOperationException. I have a new philosophy when it comes to exceptions: Don't catch them unless you can actually do something with the result (if nothing can do something with the result, then the process terminates and that's okay).

I think in our example code we shouldn't put try-catch but instead a comment mentioning it can throw various things (with the exception of InvalidCiphertextException).

README.md
```diff
@@ -13,6 +13,159 @@ HMAC-SHA256 (Encrypt-then-Mac). HKDF is used to split the user-provided key into
...
+
+### Decrypting Strings
+
+If encryption made sense, then the decryption API should be intuitive and
+precisely what you expect it to be:
+
+```php
+try {
+    $plaintext = \Defuse\Crypto\Crypto::decrypt($ciphertext, $key);
+} catch (\Defuse\Crypto\Exception\CryptoTestFailedException $ex) {
```
defuse: Again there are more exception types that can be thrown, but let's just mention that in a comment instead of trying to try...catch them all. We *do* want to try...catch the InvalidCiphertextException one in the example code though.

README.md
```diff
@@ -13,6 +13,159 @@ HMAC-SHA256 (Encrypt-then-Mac). HKDF is used to split the user-provided key into
...
+
+### Interlude: A Complete Example
+
+First, generate a key and store it:
+
+```php
+<?php
+require_once "/path/to/defuse/php-encryption/autoload.php';
+$key = \Defuse\Crypto\Key::createNewRandomKey();
+file_put_contents('shared_key.txt', $key->saveToAsciiSafeString());
```
defuse: Make this an absolute path, e.g. `/etc/examplekey.txt` just so if someone copy/pastes this it isn't saving keys to the web root by default.

defuse: Oh, I see, it's a command line application. That isn't clear until you read the paragraph below.

README.md
```diff
@@ -13,6 +13,159 @@ HMAC-SHA256 (Encrypt-then-Mac). HKDF is used to split the user-provided key into
...
+$keyData = file_get_contents('shared_key.txt');
+$key = Key::LoadFromAsciiSafeString();
+
+$decrypted = Crypto::decrypt($argv[1], $key);
+echo $decrypted, "\n";
+```
+
+If you run this command:
+
+    php decrypt_msg.php `php encrypt_msg.php It\ Works\!`
```
defuse: Maybe make decrypt read from standard in so that the command becomes:

```
echo "It works!" | php encrypt_msg.php | php decrypt_msg.php
```

Doesn't really matter to me though.

README.md
```diff
@@ -13,6 +13,159 @@ HMAC-SHA256 (Encrypt-then-Mac). HKDF is used to split the user-provided key into
...
+
+It will print "It works!" into your console. Now, assuming you and your 
+recipient have the same `shared-key.txt`, you can send messages to/from them and
+only they should be able to decrypt them.
```
defuse: Should be: "...assuming you and your recipient have the same `shared-key.txt`, **and nobody else knows the contents of that file,** you can send messages to/from them and only they **will** be able to decrypt them."

README.md
defuse: I'd fix it by making the key path a command-line option (for all the scripts). That kills two birds with one stone. It 1. Makes it obvious that it's a command-line program, and 2. Fixes the issue of people copy/pasting it into web code.

README.md
paragonie-scott: That's a good point. Don't enable "copy-paste programmers" to have a silent failure they didn't consciously write themselves.

defuse:
I'm pulling this into a branch to work on documentation.

defuse:
Oh, it's already on a branch. I'll just commit there :)

paragonie-scott:
:+1:

paragonie-scott:
Is this superceded by #226?

defuse:
Not yet, I'm probably still going to use a lot of it.

defuse:
I looked at this a lot while doing #226. I think that PR includes (at least a rewrite of) everything that was in this one.

PR #191, opened 2016-04-03
Add random_compat

### LINKED ISSUE #68: Use random_bytes() for key generation!
For PHP 7.0.0 and above, [`random_bytes()`](https://secure.php.net/random_bytes) should be used instead of [`openssl_random_pseudo_bytes()`](https://github.com/paragonie/random_compat/issues/5) (see issue).

We've written a polyfill called [random_compat](https://github.com/paragonie/random_compat) which should polyfill both APIs here. Would it make sense to consolidate all of the CSPRNG effort into that project instead and just add it as a dependency?

paragonie-scott:
We're explicitly NOT using composer for this because we need it to land in the .phar

Fixes #68

### DIFF
--- a/src/random_compat/byte_safe_strings.php
+++ b/src/random_compat/byte_safe_strings.php
@@ -1,0 +1,181 @@
+<?php
+/**
+ * Random_* Compatibility Library
+ * for using the new PHP 7 random_* API in PHP 5 projects
+ *
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2015 Paragon Initiative Enterprises
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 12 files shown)


paragonie-scott:
The only question I have is: Do we want it to fallback to `openssl_random_pseudo_bytes()`?

I'd vote no.

sstok:
> We're explicitly NOT using composer for this because we need it to land in the .phar

You can actually use Composer to download the dependency, and then using a Phar stub you make it part of the Phar as normal. This is what is done for many php phar-only projects.

paragonie-scott:
Oh, neat. I'll have to see how that's done then.

defuse:
It could still be nice to include it so that the "clone the repository and then require autoload.php" works. Using composer to fetch random_compat would remove that option, leaving only 1. use composer, or 2. use the .phar.

Actually, I'm totally fine with that. 2 options for how to use this library is simpler than 3!

src/random_compat/byte_safe_strings.php
```diff
@@ -0,0 +1,181 @@
...
+         *
+         * @param string $binary_string
+         * @param int $start
+         * @param int $length (optional)
+         *
+         * @throws TypeError
+         *
+         * @return string
+         */
+        function RandomCompat_substr($binary_string, $start, $length = null)
```
defuse: This'll have to be updated with the bugfixes, I think.

defuse: Well, actually, it doesn't, because random_compat probably never triggers the bug.

defuse:
Hang on, isn't `mcrypt_create_iv()` available even when you don't actually have mcrypt installed? If that's the case then we should just switch back to using that.

defuse:
Meh, let's just bring in random_compat then we don't have to worry about it.

defuse:
I merged this instead of waiting on @paragonie-scott to add it as a composer dependency because (1) I don't expect it to change much, and (2) I won't have to have/use composer to build the .phar.

PR #197, opened 2016-04-04
Make file and crypto compatible and lots of other stuff

defuse:

PR #211, opened 2016-04-06
Make file crypto and compatible and other things

defuse:
Supersedes #197


defuse:
Ooops I opened this against the wrong branch.

PR #212, opened 2016-04-06
Make file and crypto compatible and lots of other stuff

defuse:
Supersedes #197.


defuse:
I'm leaving this PR open until Friday in order to accept review/comments.

src/Core.php
```diff
@@ -24,11 +38,11 @@
      *
      * @return string (raw binary)
      */
-    public static function incrementCounter($ctr, $inc, &$config)
+    public static function incrementCounter($ctr, $inc, $cipherMethod)
     {
         static $ivsize = null;
         if ($ivsize === null) {
-            $ivsize = self::cipherIvLength($config->cipherMethod());
+            $ivsize = self::cipherIvLength($cipherMethod);
```
defuse: We only use this for the new format so this should hard-code `Core::CIPHER_METHOD`. I think a future commit does that but I need to verify.

src/File.php
```diff
@@ -302,21 +288,23 @@ public static function encryptResource($inputHandle, $outputHandle, Key $key)
             } else {
                 $read = self::readBytes(
                     $inputHandle,
-                    $config->bufferByteSize()
+                    Core::BUFFER_BYTE_SIZE
                 );
             }
-            $thisIv = Core::incrementCounter($thisIv, $inc, $config);
```
defuse: @paragonie-scott: Do you recall if there was a reason we were incrementing the IV before the call to `openssl_encrypt` instead of after? The two implementations are compatible after I moved this line to down below but I want to be extremely cautious about this change.

src/File.php
```diff
@@ -277,12 +263,12 @@ public static function encryptResource($inputHandle, $outputHandle, Key $key)
          * How much do we increase the counter after each buffered encryption to
```
defuse: Just above this line (github won't let me add a comment) there's another comment that says "We operate on $thisIv using a hash-based PRF derived from the initial IV for the first block." I have no idea what that means.

src/File.php
```diff
@@ -277,12 +263,12 @@ public static function encryptResource($inputHandle, $outputHandle, Key $key)
...
          * prevent nonce reuse?
          */
-        $inc = $config->bufferByteSize() / $config->blockByteSize();
+        $inc = Core::BUFFER_BYTE_SIZE / Core::BLOCK_BYTE_SIZE;
```
defuse: This scares me because if those constants aren't right, this could be a floating point number, and *who knows* what incrementCounter would do with that. (I looked, and it checks with `is_int()` so it *should* catch that, but still scary.)

src/File.php
```diff
@@ -302,21 +288,23 @@ public static function encryptResource($inputHandle, $outputHandle, Key $key)
...
                 $read,
-                $config->cipherMethod(),
+                Core::CIPHER_METHOD,
                 $ekey,
                 OPENSSL_RAW_DATA,
                 $thisIv
             );
+
+            $thisIv = Core::incrementCounter($thisIv, $inc, Core::CIPHER_METHOD);
+
```
defuse: Why aren't we just doing:
Non-exploitable sorta-bug: In the last iteration of this loop, the length of `$read` can be less than `Core::BUFFER_BYTE_SIZE` and therefore `$thisIv` is incremented too much. As an alternative we could compute `$inc` from `Core::ourStrlen($read)`.

src/File.php
defuse: We should add an assert that `$read` is less than `$inc` blocks. It's implied by the code above but it never hurts to double check something so important.

defuse:
47ed7f5edca218a71bdfc75409654317ef271bd0 LGTM

defuse:
ce7e391210a520c6f1aea92fd2575a1a4e68549a LGTM

test/unit/PasswordTest.php
```diff
@@ -1,34 +1,28 @@
...
      */
-    public function testStringInsteadOfSaltObject()
+    public function testKeyFromPasswordWrong()
     {
-        Key::CreateKeyBasedOnPassword("password", str_repeat("A", 32));
+        $pkey = KeyProtectedByPassword::CreateRandomPasswordProtectedKey("rightpassword");
+        $key1 = $pkey->unlockKey("wrongpassword");
     }
+
+    // TODO more tests (of the checksummed encoding, etc.)
```
defuse: BLOCKER: These tests need to be written.

src/KeyProtectedByPassword.php
```diff
@@ -0,0 +1,78 @@
...
+        );
+        return Key::LoadFromAsciiSafeString($inner_key_encoded);
+    }
+
+    private function __construct($salt, $encrypted_key)
+    {
+        $this->salt = $salt;
+        $this->encrypted_key = $encrypted_key;
+    }
+}
```
defuse: (Skipping review of this code since I know I rewrite it in a future commit.)

defuse:
b062787fdfcb975a4d2dde1311b5b5182ffc5a - already reviewed in its PR.

defuse:
8779ccc76aebe413094b0c4ace54c89cb2e1ce70 LGTM

src/DerivedKeys.php
```diff
@@ -0,0 +1,32 @@
...
+
+final class DerivedKeys
+{
+    private $akey = null;
+    private $ekey = null;
+
+    public function getSalt()
+    {
+        return $this->salt;
+    }
```
defuse: BLOCKER: There's no more `$this-salt`! I forgot to delete this.

src/File.php
```diff
@@ -214,24 +241,9 @@ public static function encryptResource($inputHandle, $outputHandle, Key $key)
...
-        $akey = Core::HKDF(
-            Core::HASH_FUNCTION_NAME,
-            $key->getRawBytes(),
-            Core::KEY_BYTE_SIZE,
-            Core::AUTHENTICATION_INFO_STRING,
-            $file_salt
-        );
+        $keys = $secret->deriveKeys($file_salt);
+        $ekey = $keys->getEncryptionKey();
+        $akey = $keys->getAuthenticationKey();
```
defuse: +1 for deduplicating the HKDF stuff. :)

src/File.php
paragonie-scott: Heh, yeah, once upon a time I was using CBC mode and made a kludge for that.

src/KeyOrPassword.php
```diff
@@ -0,0 +1,88 @@
+<?php
+
+namespace Defuse\Crypto;
+
+use Defuse\Crypto\Exception as Ex;
+
+final class KeyOrPassword
+{
+    const PBKDF2_ITERATIONS    = 100000;
```
defuse: Should this be increased? On my [older AMD CPU](https://defuse.ca/php-pbkdf2.htm) this takes ~0.5s using the PHP implementation. How available is the built-in one? If it's available to most users then we should tune it for the built-in one not the PHP one.

paragonie-scott: NCC Group recommended 86,000 along with their recommended bcrypt cost of 11. Since PHP defaults to a bcrypt cost of 10, I think 100,000 is fine.

src/KeyOrPassword.php
```diff
@@ -0,0 +1,88 @@
...
+        } elseif ($this->secret_type === SECRET_TYPE_PASSWORD) {
+            $prekey = Core::pbkdf2(
+                'sha256',
+                $this->secret,
+                $salt,
+                self::PBKDF2_ITERATIONS,
+                Core::KEY_BYTE_SIZE,
+                true
+            );
+            // TODO: Is reusing the same $salt between PBKDF2 and HKDF acceptable?
```
defuse: Opened #218 to ask this.

src/KeyOrPassword.php
```diff
@@ -0,0 +1,88 @@
...
+                Core::HASH_FUNCTION_NAME,
+                $this->secret->getRawBytes(),
+                Core::KEY_BYTE_SIZE,
+                Core::ENCRYPTION_INFO_STRING,
+                $salt
+            );
+            return new DerivedKeys($akey, $ekey);
+        } elseif ($this->secret_type === SECRET_TYPE_PASSWORD) {
+            $prekey = Core::pbkdf2(
+                'sha256',
```
defuse: BLOCKER: This should be `Core::HASH_FUNCTION_NAME`.

src/KeyOrPassword.php
```diff
@@ -0,0 +1,88 @@
...
+            return new DerivedKeys($akey, $ekey);
+        } else {
+            throw new Ex\CannotPerformOperationException("Bad secret type.");
+        }
+    }
+
+    public static function deriveKeysFromPassword($password)
+    {
+        $salt = Core::secureRandom(Core::SALT_BYTE_SIZE);
+    }
```
defuse: BLOCKER: Wtf is this?

src/File.php
paragonie-scott: If someone changes the constants, they're forking the library into one with different security guarantees *anyway*.

defuse:
4185778e47a011c7960c3674f75874136216b303 LGTM

src/Crypto.php
```diff
@@ -27,49 +27,47 @@ public static function createNewRandomKey()
...
      *
      * @throws Ex\CannotPerformOperationException
      * @throws Ex\CryptoTestFailedException
      *
      * @return string
      */
-    public static function encrypt($plaintext, $key, $raw_binary = false)
+    public static function encrypt($plaintext, Key $key, $raw_binary = false)
     {
-        RuntimeTests::runtimeTest();
```
defuse: **BLOCKER:** Re-add the runtime tests.

src/Crypto.php
```diff
@@ -103,18 +101,34 @@ public static function encrypt($plaintext, $key, $raw_binary = false)
      *
      * @return string
      */
-    public static function decrypt($ciphertext, $key, $raw_binary = false)
+    public static function decrypt($ciphertext, Key $key, $raw_binary = false)
     {
-        RuntimeTests::runtimeTest();
-
-        /* Attempt to validate that the key was generated safely. */
```
defuse: **BLOCKER:** Re-add the runtime tests (again).

defuse:
15c85ef6e091e3d1d647bc96734e49d74883dc36 LGTM

src/File.php
paragonie-scott: (It goes without saying: This comment is wrong and should be removed.)

defuse:
46b47c4c7708fd5d362aa1a2ebf9254803e0a114 LGTM

defuse:
4e19f42727c0ea3d33528538effeec93ef67a90b LGTM

src/KeyOrPassword.php
paragonie-scott: It appears to be an partially implemented feature.

src/File.php
defuse: Yeah, I also like security properties to be locally verifiable. I want to be able to know this code won't do something dangerous without having to open up a different file and checking that one is divisible by another. We could assert that `Core::BUFFER_BYTE_SIZE % Core::BLOCK_BYTE_SIZE === 0` here just to be extra safe.

src/Core.php
```diff
@@ -6,21 +6,21 @@
...
-    const BLOCK_BYTE_SIZE            = 16;
-    const KEY_BYTE_SIZE              = 32;
-    const SALT_BYTE_SIZE             = 32;
-    const MAC_BYTE_SIZE              = 32;
-    const HASH_FUNCTION_NAME         = 'sha256';
-    const ENCRYPTION_INFO_STRING     = 'DefusePHP|V2|KeyForEncryption';
-    const AUTHENTICATION_INFO_STRING = 'DefusePHP|V2|KeyForAuthentication';
-    const BUFFER_BYTE_SIZE           = 1048576;
+    const HEADER_VERSION_SIZE               = 4;  /* This must never change. */
+    const MINIMUM_CIPHERTEXT_SIZE           = 84; /* Absolute minimum */
```
defuse: Perhaps rename this to `CIPHERTEXT_OVERHEAD` since with CTR mode all ciphertexts are exactly 84 bytes longer than their plaintext.

defuse:
f6238bf82f910bc764db43f9ee39978acc0845d0 LGTM

defuse:
ec65415efbfde9aa0e164d073471a1b6a3b0b5aa and 5d6d96555a2a6d0f552450339e5fdec03bec0afd LGTM

defuse:
c23316ccfeb86c61d5e18e7fdc1e9cf951f60f8d LGTM

defuse:
95a5084e0785b2a73397794eacbea75733ab5b9e LGTM

defuse:
567fde674d93c74b5e73a9d7cc29d5236e4e0538 LGTM

src/File.php
```diff
@@ -58,9 +58,9 @@ private static function encryptFileInternal($inputFilename, $outputFilename, Key
             /**
              * Input file handle
              */
-            $if = \fopen($inputFilename, 'rb');
+            $if = @\fopen($inputFilename, 'rb');
```
defuse: BLOCKER: Include data from `error_get_last()` into the exception.

src/Crypto.php
defuse: Oh, they weren't ever removed, they were just moved to encryptInternal / decryptInternal.

defuse:
Blockers resolved in 0ca679d1550c9a18ed713430af9bdf4a78b94211 so once I see tests passing this is ready go go :)

PR #213, opened 2016-04-06
Re-use the stream implementation for Crypto.

defuse:
Merge this before #211


defuse:
How does this commit add 4 more lines than it removes? The whole point was to halve the codebase size! Well, I had to keep `plainEncrypt()` because it's used for the AES test vectors. I had to keep `plainDecrypt()` and because `legacyDecrypt()` uses it. It's true that `encrypt()` and `decrypt()` got removed, but they were replaced by *even longer* code for writing the strings to the streams and reading them back again.

defuse:
I'm considering rejecting this (again) because it doesn't actually make the code simpler. I'm curious to hear what @twistor and @paragonie-scott think first.

defuse:
I'm leaving this PR open until Friday in order to accept review/comments.

defuse:
Decided to reject this.

PR #226, opened 2016-04-10
Version 2.0.0 Documentation (and other things)

defuse:


defuse:
(The tests are failing because `pgp.mit.edu` is being retarded.)

larowlan:
Should/could we add a section on upgrading from 1.2?
If 2.0 is the LTS version then getting people off the earlier version will lead to lower support overhead

defuse:
@larowlan Absolutely yes! Good call.

defuse:
Finally done and ready for review. 99% chance I made a copy/paste error somewhere. The stuff in `docs/classes` is like 80% copy-pasting the same paragraph over and over so it's likely I made mistakes there.

defuse:
Note that this PR contains security-related changes, e.g. adding poor-man's domain separation for #240

docs/Tutorial.md
```diff
@@ -0,0 +1,287 @@
...
+### Scenario #1: Keep data secret from the database administrator
+
+In this scenario, our threat model is as follows. Alice is a server
+administrator responsible for managing a trusted web server. Eve is a database
+administrator responsibile for managing a database server. Dave is a web
+developer working on code that will eventually run on the trusted web server.
+
+Let's say Alice and Dave trust each other, and Alice is going to host Dave's
+application on her server. But both Alice and Dave don't trust Eve. They know
+Eve is a good database administrator, but he might have incentive to steal the
```
paragonie-scott: Eve is a guy?

defuse: Good catch! (I originally had "Mallory" and did a string replacement for "Eve").

glensc:
[you say](https://github.com/defuse/php-encryption/blob/v2-add-documentation/README.md#dependencies):
> This library requres no special dependencies except for a version of PHP 5.4 or newer with the OpenSSL extensions enabled (the default). It comes bundled with random_compat so that your users will not need to follow any special installation steps.

why it's necessary to bundle copy for third-party library in this repo, why not keep repo for *source code* only. i.e use composer dependency and for .phar files use github release page to upload the phar. repo history is already 15MiB for a library that is 56KiB size!

bundling copy of some library is directly against having secure system. if it was *not* bundled, i could update the library separately. now i install this library and don't even know maybe it has old version of third party library.

besides the solutions how not to bundle already exists, which i mentioned (composer, release uploads)

Issue #252, opened 2016-04-24
Doc Review: docs/Tutorial.md


defuse:
Review `docs/Tutorial.md` for correctness as well as understandability.

larowlan:
nit:
``` catch (Defuse\Crypto\WrongKeyOrModifiedCiphertextException) {``` is missing the ```$e``` argument

larowlan:
great docs, awesome job

defuse:
Done, made some minor fixes.

PR #254, opened 2016-04-25
Fix sample code

larowlan:


larowlan:
Fix for #226 as per https://github.com/defuse/php-encryption/issues/252#issuecomment-214553084

defuse:
Thanks!

Issue #257, opened 2016-05-09
Different way of uploading releases


defuse:
See this comment: https://github.com/defuse/php-encryption/pull/226#issuecomment-215970366

(I wasn't aware that GitHub let you upload your own files for releases.)

defuse:
@glensc If you would like to see this happen and have spare cycles to donate, it would be awesome if you could take a shot at adding `random_compat` as a composer dependency and modifying the `.phar` building script to pull `random_compat` down using compser.

glensc:
the PR should be against [v2.0](https://github.com/defuse/php-encryption/tree/v2.0) branch right?

defuse:
@glensc Yep!

glensc:
and you're sure you are against using box.phar to build?

defuse:
I'm not opposed to `box.phar` since I don't know what it is yet, but hopefully it will be easy to look at it and make sure it's generating a `.phar` that accurately represents the source code. Using it is an extra window for a backdoor / vulnerability to get inserted.

defuse:
Done in #264

paragonie-scott:
#264 is closed, so I'm closing this too.

PR #258, opened 2016-05-11
WIP: create .phar using existing tools (composer, box.phar)

glensc:
#257

WIP: basic concept, when accepted needs cleanups (dropping library copies and unused scripts)

- [x] create phar using `box.phar`
- [ ] compression to use for phar (currently GZ)
- [ ] sign phar
- [ ] drop library copies
- [ ] drop unused scripts

INSTRUCTIONS:

commands must be run  in `dist/` subdir, git tree must be clean (uncommitted changes will not be "seen")
 - to make .phar, type `make`
 - to cleanup, type `make clean`

<del>
build system will attempt to find tools from `$PATH` fallbacking to current dir, if these are missing they can be downloaded with:
 - `make composer.phar`
 - `make box.phar`
</del>


Makefile
```diff
@@ -0,0 +1,26 @@
...
+	$(composer) install --no-dev
+
+%.phar: Makefile box.json composer.lock
+	$(php) -d phar.readonly=0 $(box) build -v
+
+box.phar:
+	curl -LSs https://box-project.github.io/box2/installer.php | php
+
+composer.phar:
+	curl -sS https://getcomposer.org/installer | php
```
defuse: NACK on the above two curl commands. Let's instead document that developers need to install box.phar and composer in `docs/InternalDeveloperDocs`, and let them decide how securely they want to acquire them.

Makefile
```diff
@@ -0,0 +1,26 @@
+TARGETS := defuse-crypto.phar
```
defuse: Can we put things related to `.phar` building like this inside `other/`? I don't want regular users to get confused and think they need to run make. Building the `.phar` is something only this library's developers should do (everyone else should be using a signed `.phar`).

Makefile
glensc:
```
➔ PATH=/usr/bin make
Makefile:7: *** Required tool `box' not installed, see docs/InstallingAndVerifying.md.  Stop.
```

glensc:
move the dist making scripts to dist/ subdir, also dropped automatic download and point use to documentation if tool can not be found from local system.

defuse:
I incorporated this into #264, thanks!

PR #267, opened 2016-05-17
Remove Key::loadFromRawBytesForTestingPurposesOnlyInsecure

mjrider:
the function loadFromRawBytesForTestingPurposesOnlyInsecure should only be used for unit testing. 

this PR removes the internal function and uses a closure to create a Key instance with the right key material.

### DIFF
--- a/src/Key.php
+++ b/src/Key.php
@@ -81,17 +81,4 @@ final class Key
         $this->key_bytes = $bytes;
     }
 
-    /**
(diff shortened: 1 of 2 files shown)


defuse:
I like this change a lot! Thanks!

sstok:
Genius 😵

glensc:
just saying to confirm this matches php min version 5.4.0 in composer.json
http://php.net/manual/en/closure.bindto.php

PR #272, opened 2016-05-22
Don't export test for Composer package

laurentj:
Downloading php-encryption via Composer takes too much time for a 232KB library. This is because tests, having 13 MB of files, are including in the package. It is very inconvenient in CI environments or for user of projects that use php-encryption, as it takes time, band-width etc..

A user of the library doesn't need theses tests.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -2,4 +2,3 @@
 /.travis.yml export-ignore
 /test.sh export-ignore
 /test export-ignore
-/dist export-ignore
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,5 @@
 /.gitattributes export-ignore
 /.travis.yml export-ignore
+/test.sh export-ignore
+/test export-ignore
+/dist export-ignore


sethbattin:
Disagree with this one.  The test code should be available with the library code, regardless of distribution method.  Any code consumer can elect to exclude the test sources if they feel that it's import to their deployment or other CI.  But if I met that type of consumer, I would tell them they are focusing on the wrong kind of optimization.

laurentj:
>Any code consumer can elect to exclude the test sources if they feel that it's import to their deployment or other CI.

We cannot say to Composer to exclude some parts of a package. But we can say to Composer to not download the package, and download directly the git content.

For a user, most of time, tests are totally useless. He just want to use the library. It is ridiculous to download 13MB of files that are useless. It represents **98%** of the package! 98% of useless files. If he want tests with Composer (most of time, when he wants to contribute), he can say to Composer to download the git content of the package, or he can download an other kind of package (but who download a package to contribute? most of time a contributor clones the git repo)  

Note also that it consumes uselessly band-width on packagist.org.

sstok:
If you need the source simple use `--prefer-source` and remove the vendor folder to ensure it's downloaded. But most users will never run the tests of there vendor (as they will actually break due to missing dev dependencies).

And the runtime tests are already done during the usage of the library itself 👍

sethbattin:
Fair enough.  If it's still trivial to fetch the full repo, no harm from minimizing it as the default.

I still disagree that this is a worthwhile optimization.  If you want to save bandwidth, don't dynamically download packages at test run time.  If you're not running the library's tests, you probably want a static version of its code anyway.

sstok:
`don't dynamically download packages at test run time.`? Can you explain this one :)

php-encryption performs runtime-tests to ensure a sane crypto system, which is already part of the Core of the package itself, you can't skip/disable these. https://github.com/defuse/php-encryption/blob/606645d1826312f111c9f761c7d994d890766042/src/Crypto.php#L178

laurentj:
@sethbattin to be clear, I talked about the tests/ directory, not about RuntimeTests. All stuff in the test/ directory  are totally useless for a normal user, especially when the package is downloaded as a dependency of an other package, and not as a package specified directly by the user. 

tests/ is only useful for a contributor (and a contributor doesn't use Composer to retrieve the source code, but he uses git directly IMHO, as there is high probability that he will clone the project in Github, and will do a PR).

sethbattin:
@laurentj yes, I follow your meaning.  You convinced me that this change is fine; the tests can be run at any time.

@sstok What I mean is that you could simply stop using composer to deploy if you absolutely must save download bytes.  Just keep a local copy instead.  I think it's contradictory to both try to use composer as a per-test deployment system, and simultaneously try to limit download size.  You would save more bandwidth by checking out the full repo once than by dynamically downloading 2% of it 50 times.

PR #274, opened 2016-06-01
Properly detect mbstring overloading.

twistor:
This provides some benefits:
1. Avoids using mb_\* functions if an mbstring polyfill is used. https://github.com/symfony/polyfill-mbstring
2. Avoids the mb_substr() bug for most installs.
3. Uses the simpler/faster codepath the majority of the time.

Edit:
There were some failing tests after the change. I've decided to just move all of the bounds checking to the beginning. I think this simplifies the code significantly. And defends against new errors is substr() handling.

### DIFF
--- a/src/Core.php
+++ b/src/Core.php
@@ -265,7 +265,7 @@ final class Core
     {
         static $exists = null;
         if ($exists === null) {
-            $exists = \function_exists('mb_strlen');
+            $exists = \extension_loaded('mbstring') && \ini_get('mbstring.func_overload') & MB_OVERLOAD_STRING;
         }
         if ($exists) {
             $length = \mb_strlen($str, '8bit');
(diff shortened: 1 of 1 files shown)


sstok:
Should be save as this option can only be changed with the php.ini, so it cannot be changed at run time. http://php.net/manual/en/mbstring.configuration.php#ini.mbstring.func-overload

FYI, it's possible to add an extra job for Travis, where `mbstring.func_overload` is enabled.
So it's possible to test if it actually works 👍

defuse:
There is a bug in both the original and new code when `$length` is negative (meaning the substring should end as if the original string had `-$length` characters chopped off) and `$start` is an index into a character that would be chopped off. If `$exists` is true then the old code will go on calling `mb_substr` with a negative `$length` and it's undefined what will happen. The new code does the same thing except its `$exists` is less likely to be true. Both old and new will end up (assuming the `mb_substr` call returns) throwing an `EnvironmentIsBrokenException` because of the check for PHP bug #66797 (but in this case the error message will be incorrect).

defuse:
I'll solve the above by changing it to throw an exception when `$length` is negative.

Another discrepancy between this function and `substr` is that `substr` says "If length is given and is positive, the string returned will contain **at most length characters** beginning from start (depending on the length of string)." That means you could call it with a huge length and expect to get everything up to the end of the string. But if `$exists` is true then both old and new code will throw the exception for PHP bug #66797. If `$exists` is false it just calls `substr` so it does the right thing. So the code behaves differently depending on whether the `mb_substr` stuff is turned on.

defuse:
Thankfully `mb_substr` treats its positive-valued `$length` parameter the same way as `substr` so we can fix that by changing the bug-check to `if (Core::ourStrlen($substr) < $length)`.

defuse:
I merged this, with those fixes, in #389.

defuse:
Hang on, I'm an idiot, changing the check to `if (Core::ourStrlen($substr) < $length)` has the same problem. The check should actually check if it's 0. Fixing that over in #389...

PR #279, opened 2016-07-12
Remove large.jpg

defuse:


paragonie-scott:
Yeah, we don't need this anymore. It was good for benchmarking the file encryption (vis a vis memory usage).

PR #292, opened 2016-09-06
use raw link to gpg key

glensc:
because github renders it as html and it messes up gpg separators

image

alternative would be to rename the file to have .txt extension


sstok:
Should this problem also be reported to GitHub?

defuse:
Thanks!

PR #295, opened 2016-09-09
Add hhvm to Travis CI configuration

paragonie-scott:


paragonie-scott:
HHVM and Phar don't seem to play well together. @sgolemon is this a known issue upstream?

PR #300, opened 2016-09-21
Fix return type of Key::createNewRandomKey()

theodorejb:
Without a leading backslash the type is interpreted relative to the current namespace.

bug


defuse:
Thank you!

PR #305, opened 2016-10-08
Need to composer install in order to run tests

defuse:


defuse:
I'm making a trade-off here. There's a risk doing `composer install` could taint the release in some way, but the only thing it could do is affect the `.phar`, and I think the threat of releasing something which doesn't pass unit tests because we didn't test them (e.g. a bug in the Makefile leads to an empty `.phar` or something) is greater.

PR #309, opened 2016-10-13
Updated constraint on OpenSSL

mickaelandrieu:
Hi,

as AES-CTR-256 is declared as the default CIPHER METHOD and as it's only available since OpenSSL 1.0.1 I guess the requirements should be updated.

A lot of basic web providers, like OVH, 1&1 or Go daddy are still using OpenSLL 0.98 and can't enjoy by default your nice library.

Any idea on how to use this library with OpenSLL 0.98 ? We use it for PrestaShop and we want to ease the installation for final users.

Mickaël

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,7 @@
     },
     "require": {
         "paragonie/random_compat": "~2.0",
-        "ext-openssl": "*",
+        "ext-openssl": ">=1.0.1",
         "php":  ">=5.4.0"
     },
     "require-dev": {


paragonie-scott:
> A lot of basic web providers, like OVH, 1&1 or Go daddy are still using OpenSLL 0.98 and can't enjoy by default your nice library.

That sounds like we really need to get the OVH, 1&1, and GoDaddy security teams to take note of this and fix the problem on their end, then.

mickaelandrieu:
@paragonie-scott  this will be the ideal solution in an ideal world, I totally agree with you on this point :-) 

I'm not trolling, I'm really facing a problem that any PrestaShop user can meet when installing the solution, and they are not developers or whatever :/

I guess I could use the old `v1.2.1`, what do you think ?

Still, the requirements should be set up ^^

paragonie-scott:
Err, no. The requirement setting is actually false; [see the failing tests on 5.6 and below](https://travis-ci.org/defuse/php-encryption/builds/167304355).

This is the real problem:

> Some 0.9.8 versions of OpenSSL are in FIPS mode, which means CTR mode was removed at compile time.

See #280 

To work towards a permanent solution, I'm reaching out to the security teams at these large hosting providers right now to get their clients off of 0.9.8-fips unless they specifically need it to fulfill compliance requirements.

If you need to use v1.2.1 for the interim, it's secure but [not as good as v2](https://paragonie.com/blog/2016/05/defuse-security-s-php-encryption-library-version-2-0-0-released).

jedisct1:
Is that on OVH shared hosting? I'm gonna see what I can do.

mickaelandrieu:
@paragonie-scott @jedisct1 yes, could be nice thank you :)

jedisct1:
Response from the OVH shared hosting team (sorry, I'm not that familiar with that product yet):

Multiple images are available, and all it takes to switch to a different image is updating the ovhconfig file.

Only the "legacy" image ships with openssl 0.9.8. This is a very old environment, that some customers unfortunately still don't want to migrate from.

All other images, including the "stable" one installed by default, currently use OpenSSL 1.0.1t.

Slightly outdated reference: https://docs.ovh.com/fr/fr/web/hosting/modifier-lenvironnement-dexecution-de-mon-hebergement-web/#differences-entre-les-images

Looks like the libsodium extension might also be included in the next release of the stable image.

paragonie-scott:
@jedisct1 Thanks for getting to the bottom of that.

> Looks like the libsodium extension might also be included in the next release of the stable image.

7drhiqrh

Issue #310, opened 2016-11-17
Tutorial should explain the Defuse\Crypto namespace


defuse:
The code example in the Tutorial doesn't work, e.g.

```
$protected_key = KeyProtectedByPassword::createRandomPasswordProtectedKey($password);
```

should be

```
$protected_key = \Defuse\Crypto\KeyProtectedByPassword::createRandomPasswordProtectedKey($password);
```

defuse:
It would be nice if the code in the markdown files could be tested automatically somehow. Someone ought to have thought of testable documentation already... so are there any good tools for that out there?

timeglider:
Here here! I pulled a lot of hair off my head before I found this:
$key = \Defuse\Crypto\Key::createNewRandomKey();
instead of 
$key = Key::createNewRandomKey();

paragonie-scott:
The tutorial has been updated. Is it better now?

paragonie-scott:
It's been a few months since this was fixed. I hope the fix was satisfactory. I'm going to close the issue but feel free to correct me if there's still anything left to do and I will reopen it.

PR #311, opened 2016-11-19
Add use declarations to Tutorial

paragonie-scott:
See #310

### DIFF
--- a/docs/Tutorial.md
+++ b/docs/Tutorial.md
@@ -105,9 +105,11 @@ encryption key and prints it to standard output:
 
 ```php
 <?php
-    // generate-key.php
-    $key = Key::createNewRandomKey();
-    echo $key->saveToAsciiSafeString();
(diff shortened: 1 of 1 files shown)


paragonie-scott:
Note: in Github flavored markdown, if you demarcate your code blocks with triple backticks, you don't need to indent the interiors.

docs/Tutorial.md
```diff
@@ -215,63 +225,67 @@ logged in during the attack.
...
+// ... when the user is logging out ...
+// ... securely wipe the saved $user_key_encoded from the system ...
 ```
 
 When a user adds their credit card number, Dave's code will get the key from the
 session and use it to encrypt the credit card number:
 
 ```php
 <?php
+use Defuse\Crypto\Key;
```
defuse: I think this needs Defuse\Crypto\Crypto, too.

Issue #315, opened 2017-01-03
Integrity check failed


iyasilias:
I encrypt this string **Dato Farizan 150.jpg**, works fine.

But to decrypt the encrypted string, I get **Fatal error**: Uncaught exception 'Defuse\Crypto\Exception\WrongKeyOrModifiedCiphertextException' with message 'Integrity check failed.'

At first I thought it is due to the space in the string, but I tried **DatoFarizan150.jpg**, still the problem persists.

Please help.

iyasilias:
Sorry my bad.

Forget to mention that I've stored the encrypted string in a database. And the reason I get the error is because I use varchar(255) to store the encrypted string, so the encrypted string not fully stored.

So, do you know exactly or estimably how long the encrypted string will be?

paragonie-security:
> So, do you know exactly or estimably how long the encrypted string will be?

```php
$message = Defuse\Crypto\Crypto::encrypt($yourMessage, $yourKey);
var_dump(strlen($message));
```

jimmykane:
@paragonie-scott Someone could have guessed that the length is variable.

paragonie-security:
That's the easiest way to estimate it. (Also, maybe use `TEXT` instead of `VARCHAR(255)`.)

iyasilias:
> Also, maybe use TEXT instead of VARCHAR(255)
Yes that's my solution for now.

Issue #56, opened 2015-05-15
Best Mysql field type to store encrypted data


bab55z:
Hello!
I am wondering about what is the best type i should choose for the field i use to store the encrypted data, i chose varchar but encrypted data seems to be corrupted after it is stored in mysql. i Planing to use a blob type, is this a good choice and what size should choose also. I planning to encrypt integers only which should not exceed 8 or 9 digits,
my Mysql database uses innoDB engine with utf8_general_ci

thanks

paragonie-scott:
MySQL tends to munge everything, so I would definitely `base64_encode()` it before storing and `base64_decode()` after retrieving.

bab55z:
works great! thanks @paragonie-scott

goldsky:
this should be in the doc

paragonie-scott:
@goldsky That's a fair point. We're going to be discussing documentation in #52 if you would like to follow that discussion and weigh in with your own opinions.

PR #316, opened 2017-01-03
Add getExpectedCiphertextLength() helper method

paragonie-security:
See #315 #56 and possibly others

### DIFF
--- a/src/Crypto.php
+++ b/src/Crypto.php
@@ -163,6 +163,36 @@ class Crypto
             );
         }
     }
+    
+    /**
+     * Calculate the expected string length of a ciphertext given
+     * the length of a string.
+     *
+     * @param int $integerContainingStringLength
+     * @param bool $rawOutput
+     * @return int
+     */
+    public static function getExpectedCiphertextLength($integerContainingStringLength, $rawOutput = false)
+    {
+        if (is_string($integerContainingStringLength)) {
+            throw new \InvalidArgumentException(
+                'getExpectedCiphertextLength() expects an integer, not a string. Pass strlen($var) instead of $var'
+            );
+        } elseif (!is_int($integerContainingStringLength)) {
+            throw new \InvalidArgumentException(
+                'getExpectedCiphertextLength() expects an integer'
+            );
+        }
+        $rawEstimate = $integerContainingStringLength;
+        $rawEstimate += 4;  // Header
+        $rawEstimate += 16; // Nonce for AES-CTR
+        $rawEstimate += 32; // HKDF Salt
+        $rawEstimate += 32; // MAC
+        if ($rawOutput) {
+            return (int) $rawEstimate;
+        }
+        return (int) ($rawEstimate * 2); // Hex encoded
+    }
 
     /**
      * Encrypts a string with either a key or a password.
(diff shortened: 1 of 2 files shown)


src/Crypto.php
```diff
@@ -163,6 +163,36 @@ public static function legacyDecrypt($ciphertext, $key)
...
+     * Calculate the expected string length of a ciphertext given
+     * the length of a string.
+     *
+     * @param int $integerContainingStringLength
+     * @param bool $rawOutput
+     * @return int
+     */
+    public static function getExpectedCiphertextLength($integerContainingStringLength, $rawOutput = false)
+    {
+        if (is_string($integerContainingStringLength)) {
```
paragonie-security: I can, alternatively, make this silently just do an internal mbstring.func_overload-safe `strlen()` (even though it encourages bad habits).

src/Crypto.php
```diff
@@ -163,6 +163,36 @@ public static function legacyDecrypt($ciphertext, $key)
...
+    
+    /**
+     * Calculate the expected string length of a ciphertext given
+     * the length of a string.
+     *
+     * @param int $integerContainingStringLength
+     * @param bool $rawOutput
+     * @return int
+     */
+    public static function getExpectedCiphertextLength($integerContainingStringLength, $rawOutput = false)
```
defuse: What guarantee does this provide? I can see several different meanings of "expected" which needs to be clarified:

1. 99.99% of time the returned value is exact, but 0.01% of the time it's too large by a factor of 1,000.
2. If this function returns N, and the real size of the ciphertext is R, then `R < N < 2R` or `R < N < R+C` for some constant C.
3. The returned value is always exact.

Please document precisely which it is.

src/Crypto.php
defuse: I like the current implementation.

test/unit/CryptoTest.php
```diff
@@ -84,6 +84,30 @@ public function testEncryptDecryptWithPassword()
         } catch (Ex\WrongKeyOrModifiedCiphertextException $e) { /* expected */
         }
     }
+    
+    public function testGetExpectedCiphertextLength()
+    {
+        $strings = array(
+            '',
+            'This is a test message',
+            str_repeat('Test message.' . "\n", 32)
```
defuse: We should check all 1-block and 2-block congruences modulo the block size, so 

```
"",
"a",
"aa",
...
"aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
```
i.e. `for ($i = 0; $i <= 32; $i++) { $str = str_repeat("a", $i); ...`

(Also I think this answers the question I asked in the previous comment)

src/Crypto.php
paragonie-scott: I just realized I never revisited this PR. It's door number 3, since it's just integer arithmetic.

src/Crypto.php
lookyman: Maybe you wanted `\is_string`?

src/Crypto.php
```diff
@@ -163,6 +163,36 @@ public static function legacyDecrypt($ciphertext, $key)
...
+    {
+        if (is_string($integerContainingStringLength)) {
+            throw new \InvalidArgumentException(
+                'getExpectedCiphertextLength() expects an integer, not a string. Pass strlen($var) instead of $var'
+            );
+        } elseif (!is_int($integerContainingStringLength)) {
```
lookyman: Maybe you wanted `\is_int`?

paragonie-scott:
Will revisit this later. For the time being, people can easily figure out that the ciphertext length is `$plaintextLength + 84` for raw binary and `($plaintextLength + 84) * 2` for hex-encoded strings.

PR #319, opened 2017-02-05
travis: add PHP 7.1

TomasVotruba:


defuse:
Let's see if #320 fixes the failing tests

PR #320, opened 2017-02-06
Use explicit version numbers in PHPUnit URLs

defuse:


TomasVotruba:
The standard place to put PHPUnit is `composer.json`

Why not put it there? It allows people to test locally with whatever the PHP version is.

defuse:
@TomasVotruba: Good point, I think a long time ago before we used composer, it was for security reasons so that we could check the cryptographic signatures ourselves. But I can't remember the rational, so I opened #322.

TomasVotruba:
@defuse Yea, I saw that. That looked like so much effort just to download one package.

Thank for opening the issue. I can send PR for this if needed.

PR #321, opened 2017-02-06
WIP: composer: use PSR-4 autoload

TomasVotruba:
I think you might use PSR-4 autoloading. Classes are in correct places.


TomasVotruba:
Glad to help. Tests are passing, to yes!

TomasVotruba:
Thanks

PR #326, opened 2017-02-26
Another example project to Readme

tsusanka:
I wrote a simple file encrypt/decrypt example [fileencrypt](https://github.com/tsusanka/fileencrypt), quite similar to yours [encutil](https://github.com/defuse/encutil).

It uses Symfony Console interface so it can have a litte bit of additional value. Feel free to include it in your Readme if desired.


defuse:
Thank you!

PR #332, opened 2017-03-24
Fix for issue 330

### LINKED ISSUE #330: Using hash_hkdf() for PHP 7.1.2 and above
Hello,  
First of all, thanks for this great library!  
I’m tinkering with the code trying to simplify ParagonIe’s [PasswordLock](https://github.com/paragonie/password_lock) to get everything, including password hashing and authenticated encryption, to one single file without breaking the functionality and security.  
And, when looking at the `core::HKDF()` method, I found out that since PHP 7.1.2 there is a built-in [hash_hkdf](http://php.net/manual/en/function.hash-hkdf.php) method.  
Is it worth to use it with appropriate checks? I think so, but just decided to ask before making a PR.  
Thanks!

Oire:
Fixes #330: using \hash_hkdf() when possible

### DIFF
--- a/src/Core.php
+++ b/src/Core.php
@@ -129,6 +129,14 @@ final class Core
      */
     public static function HKDF($hash, $ikm, $length, $info = '', $salt = null)
     {
+        static $nativeHKDF = null;
+        if ($nativeHKDF === null) {
+            $nativeHKDF = \function_exists('hash_hkdf');
+        }
+        if ($nativeHKDF) {
+            return \hash_hkdf($hash, $ikm, $length, $info, $salt);
+        }
+
         $digest_length = Core::ourStrlen(\hash_hmac($hash, '', '', true));
 
         // Sanity-check the desired output length.


src/Core.php
```diff
@@ -129,6 +129,14 @@ public static function secureRandom($octets)
      */
     public static function HKDF($hash, $ikm, $length, $info = '', $salt = null)
     {
+        static $nativeHKDF = null;
+        if ($nativeHKDF === null) {
+            $nativeHKDF = \function_exists('hash_hkdf');
```
paragonie-scott: We typically use `is_callable()` (because Suhosin is silly), and it may be worth prefixing the function name `\\` just to be on the safe side.

paragonie-scott: Actually, scratch that. That change never landed in this library, only random_compat. I'll fix that before the next release.

Oire: Oh, I'm still learning, I took `hashEquals()` as a template.  
If you think that `is_callable()` is more... what? trustful? I need some Googling about Suhosin issues, I guess, — I could send another PR to fix that through all of the library code (since I've been digging it quite thoroughly, it won’t take much time).

paragonie-scott: No, it's identical.

Suhosin is a security patch set that the OpenBSD team bundles with PHP, and it does a lot of stupid things like disabling `function_exists()`.

paragonie-scott: Don't worry about this, I'll take care of it tonight. :)

@defuse - this PR looks good. Any objections to merging?

defuse: LGTM

Issue #334, opened 2017-04-10
A little help please?


HawkonDK:
Uncaught Defuse\Crypto\Exception\BadFormatException: Encoding::hexToBin() input is not a hex string. in phar:///Users/user/Downloads/encrypt/defuse-crypto.phar/src/Encoding.php:65 Stack trace: #0 phar:///Users/user/Downloads/encrypt/defuse-crypto.phar/src/Encoding.php(164): Defuse\Crypto\Encoding::hexToBin('keys.txt') #1 phar:///Users/user/Downloads/encrypt/defuse-crypto.phar/src/Key.php(38): Defuse\Crypto\Encoding::loadBytesFromChecksummedAsciiSafeString('\xDE\xF0\x00\x00', 'keys.txt') #2 /Users/user/Downloads/encrypt/test.php(19): Defuse\Crypto\Key::loadFromAsciiSafeString('keys.txt') #3 /Users/user/Downloads/encrypt/test.php(24): loadEncryptionKeyFromConfig() #4 {main} thrown in phar:///Users/user/Downloads/encrypt/defuse-crypto.phar/src/Encoding.php on line 65

Note: I have removed my own name by /user/

paragonie-scott:
This appears to be a duplicate of #333. Can you please check and see if that solves your problem?

HawkonDK:
nope didn't help but find a way myself ;)

defuse:
Oh, it looks like it was because the actual string `'keys.txt'` was being passed to `loadFromAsciiSafeString`. Yeah, that function expects the actual key string, not a file path.

Issue #333, opened 2017-04-05
Fatal error: Uncaught exception


gavinhwatson:
Hi Guys,

Apologies is this is the wrong forum for this...

I have the library running perfectly locally Windows, xampp... but when uploading everything to a Ubuntu,xampp server I get the following error when trying to use it...

Fatal error: Uncaught exception 'Defuse\Crypto\Exception\BadFormatException' with message 'Encoding::hexToBin() input is not a hex string.' in phar:///opt/lampp/htdocs/php-encryption/defuse-crypto.phar/src/Encoding.php:65 Stack trace: #0 phar:///opt/lampp/htdocs/php-encryption/defuse-crypto.phar/src/Encoding.php(164): Defuse\Crypto\Encoding::hexToBin('def00000af5605e...') #1 phar:///opt/lampp/htdocs/php-encryption/defuse-crypto.phar/src/Key.php(38): Defuse\Crypto\Encoding::loadBytesFromChecksummedAsciiSafeString('\xDE\xF0\x00\x00', 'def00000af5605e...') #2 /opt/lampp/htdocs/projects/new_project.php(23): Defuse\Crypto\Key::loadFromAsciiSafeString('def00000af5605e...') #3 /opt/lampp/htdocs/projects/new_project.php(25): loadEncryptionKeyFromConfig() #4 {main} thrown in phar:///opt/lampp/htdocs/php-encryption/defuse-crypto.phar/src/Encoding.php on line 65

I've tried generating a new key but that still produces the exact same error.

Would really appriciate your help!

Gavin

paragonie-scott:
```
Stack trace:
#0 phar:///opt/lampp/htdocs/php-encryption/defuse-crypto.phar/src/Encoding.php(164): Defuse\Crypto\Encoding::hexToBin('def00000af5605e...')
#1 phar:///opt/lampp/htdocs/php-encryption/defuse-crypto.phar/src/Key.php(38): Defuse\Crypto\Encoding::loadBytesFromChecksummedAsciiSafeString('\xDE\xF0\x00\x00', 'def00000af5605e...')
#2 /opt/lampp/htdocs/projects/new_project.php(23): Defuse\Crypto\Key::loadFromAsciiSafeString('def00000af5605e...')
#3 /opt/lampp/htdocs/projects/new_project.php(25): loadEncryptionKeyFromConfig()
[...]

PR #335, opened 2017-04-10
Strip Newlines from ASCII-Safe Strings

paragonie-scott:
This would prevent the errors present in #333 and #334. I don't believe removing newlines in variable time would ever introduce a practical timing leak, but I'll wait until @defuse sanity-checks this.

### DIFF
--- a/src/Encoding.php
+++ b/src/Encoding.php
@@ -79,9 +79,10 @@ final class Encoding
     }
     
     /**
-     * Implement rtrim() without table lookups or branches.
-     * This is still variable-time with regards to string length, but only leaks
-     * how much whitespace was at the end of the string.
+     * Remove trialing whitespace without table look-ups or branches.
+     *
+     * Calling this function may leak the length of the string as well as the
+     * number of trailing whitespace characters through side-channels.
      *
      * @param string $string
      * @return string
@@ -89,8 +90,11 @@ final class Encoding
     public static function trimTrailingWhitespace($string = '')
     {
         $length = Core::ourStrlen($string);
+        if ($length < 1) {
+            return '';
+        }
         do {
-            $prev = $length;
+            $prevLength = $length;
             $last = $length - 1;
             $chr = \ord($string[$last]);
 
@@ -126,7 +130,7 @@ final class Encoding
             // if ($chr === 0x20) $length -= 1;
             $sub = (((0x1f - $chr) & ($chr - 0x21)) >> 8) & 1;
             $length -= $sub;
-        } while ($prev !== $length && $length > 0);
+        } while ($prevLength !== $length && $length > 0);
         return Core::ourSubstr($string, 0, $length);
     }
 
@@ -213,12 +217,9 @@ final class Encoding
             );
         }
 
-        try {
-            $bytes = Encoding::hexToBin($string);
-        } catch (Ex\BadFormatException $ex) {
-            // If this fails once, try without whitespace.
-            $bytes = Encoding::hexToBin(Encoding::trimTrailingWhitespace($string));
-        }
+        /* If you get an exception here when attempting to load from a file, first pass your
+           key to Encoding::trimTrailingWhitespace() to remove newline characters, etc.      */
+        $bytes = Encoding::hexToBin($string);
 
         /* Make sure we have enough bytes to get the version header and checksum. */
         if (Core::ourStrlen($bytes) < self::SERIALIZE_HEADER_BYTES + self::CHECKSUM_BYTE_SIZE) {
--- a/src/Encoding.php
+++ b/src/Encoding.php
@@ -77,6 +77,39 @@ final class Encoding
         }
         return $bin;
     }
+    
+    /**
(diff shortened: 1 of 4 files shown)


defuse:
It looks like `rtrim` is implemented by [looking up each character of the string in a mask table](https://github.com/php/php-src/blob/c6982995504b6e21e8a5ade29cfb16a55196dc43/ext/standard/string.c#L839) which potentially leaks the key through cache side channels. :(

I'm not sure if it's worth writing a timing-safe `trim` just for this issue... I wonder if making the error message more informative (like say what the unexpected characters are) would help people realize what's going on in most cases. 

I'm open to ideas.

paragonie-scott:
If we tell people, they'll just use rtrim themselves. As long as we're not concerned with length-based timing leaks in this specific context, I think this should be fine with our own implementation.

src/Key.php
```diff
@@ -35,6 +35,7 @@ public static function createNewRandomKey()
      */
     public static function loadFromAsciiSafeString($saved_key_string)
     {
+        $saved_key_string = Encoding::rtrim($saved_key_string, "\r\n");
```
defuse: `Encoding::rtrim` only takes one parameter -- maybe name it something else to avoid confusing it with the built-in `rtrim`.

paragonie-scott:
I renamed the function, opened this in a proper IDE, and got the logic fixed for `0x00` byte handling. I also made the unit test more useful. This should be in a state where we can consider merging it.

I added a try/catch to automatically handle these whitespace-induced errors, for the sake of enhancing usability. If we decide that's a bad idea, I'd definitely suggest implementing them inside of a `loadKeyFromFile` method.

src/Encoding.php
```diff
@@ -161,7 +213,12 @@ public static function loadBytesFromChecksummedAsciiSafeString($expected_header,
             );
         }
 
-        $bytes = Encoding::hexToBin($string);
+        try {
+            $bytes = Encoding::hexToBin($string);
+        } catch (Ex\BadFormatException $ex) {
+            // If this fails once, try without whitespace.
+            $bytes = Encoding::hexToBin(Encoding::trimTrailingWhitespace($string));
+        }
```
paragonie-scott: This is probably the only controversial part of this PR that remains: @defuse

defuse:
Thanks! I'll take a look soon.

src/Encoding.php
```diff
@@ -77,6 +77,58 @@ public static function hexToBin($hex_string)
...
+     * how much whitespace was at the end of the string.
+     *
+     * @param string $string
+     * @return string
+     */
+    public static function trimTrailingWhitespace($string = '')
+    {
+        $length = Core::ourStrlen($string);
+        do {
+            $prev = $length;
```
defuse: It would be easier to understand if this was called `$prev_length`.

src/Encoding.php
```diff
@@ -77,6 +77,58 @@ public static function hexToBin($hex_string)
...
+            $last = $length - 1;
+            $chr = \ord($string[$last]);
```
defuse: If you call this on the empty string, it will try to access `$string[-1]`.

src/Encoding.php
defuse: Yeah, let's not do this here since it's an internal-only function. This function should fail if the string is not precisely what it's expected to be. It's hard to tell what this will affect (e.g. does this change allow appending newlines to ciphertexts? I can't tell without going to look at other code, so this violates "security should be locally verifiable").

src/Key.php
```diff
@@ -35,6 +35,7 @@ public static function createNewRandomKey()
      */
     public static function loadFromAsciiSafeString($saved_key_string)
     {
+        $saved_key_string = Encoding::trimTrailingWhitespace($saved_key_string);
```
defuse: We also need to update the docs (the comment above as well as the separate markdown docs) to mention that it has this new behavior.

test/unit/EncodingTest.php
```diff
@@ -82,4 +82,42 @@ public function testBadHexEncoding()
         $str[0] = 'Z';
         Encoding::loadBytesFromChecksummedAsciiSafeString($header, $str);
     }
+
+    /**
+     * This shouldn't throw an exception.
+     */
+    public function testPaddedHexEncoding()
```
defuse: This test looks good!

src/Encoding.php
paragonie-scott: That's why the loop terminates where it does, but in a do-while the condition you allude to is definitely possible.

src/Encoding.php
paragonie-scott: Understood.

paragonie-scott:
OK, I made the changes you requested.

I also added (and documented) an optional boolean parameter to `Key::loadFromAsciiSafeString()`, which if set to true, skips the whitespace trim step (enabled by default). This should allow people to pick up the library easier and side-step a lot of the minor complications people encounter frequently.

PR #342, opened 2017-04-25
[WIP] Add command line script for generating new keys

### LINKED ISSUE #338: Feature proposal: command line tool for generating a new key
In one of our projects that uses this library, we have a small PHP script `bin/generate-key.php` which simply generates a new key and prints it out:

```php
#!/usr/bin/env php
<?php

require __DIR__ . '/../vendor/autoload.php';

$key = Defuse\Crypto\Key::createNewRandomKey();
echo $key->saveToAsciiSafeString(), "\n";
```

I imagine others would have a similar need. Perhaps something like this could be added to this project as a [Composer vendor binary](https://getcomposer.org/doc/articles/vendor-binaries.md) so that one would simply need to run `vendor/bin/generate-defuse-key` (or whatever) to generate a new key.

Thoughts?

chriseskow:
Closes #338

### DIFF
--- a/bin/generate-defuse-key
+++ b/bin/generate-defuse-key
@@ -1,0 +1,14 @@
+#!/usr/bin/env php
+<?php
+
+use Defuse\Crypto\Key;
+
+foreach ([__DIR__ . '/../../../autoload.php', __DIR__ . '/../vendor/autoload.php'] as $file) {
+    if (file_exists($file)) {
+        require $file;
+        break;
+    }
+}
+
+$key = Key::createNewRandomKey();
+echo $key->saveToAsciiSafeString(), "\n";
--- a/composer.json
+++ b/composer.json
@@ -27,5 +27,8 @@
     },
     "require-dev": {
         "nikic/php-parser": "^2.0"
-    }
+    },
+    "bin": [
+        "bin/generate-defuse-key"
+    ]
 }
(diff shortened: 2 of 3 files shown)


chriseskow:
Updated the tutorial with a reference to the script.

bin/generate-defuse-key
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env php
+<?php
+
+use Defuse\Crypto\Key;
```
defuse: Doesn't the include need to happen before this statement will work?

paragonie-scott: No. Namespace imports go at the top.

paragonie-scott:
Fixed a merge conflict introduced by pulling another branch.

chriseskow:
@paragonie-scott Thanks! Anything else needed for this pull request?

bin/generate-defuse-key
```diff
@@ -0,0 +1,14 @@
...
+
+foreach ([__DIR__ . '/../../../autoload.php', __DIR__ . '/../vendor/autoload.php'] as $file) {
+    if (file_exists($file)) {
+        require $file;
+        break;
+    }
+}
```
maidmaid:
```php
require file_exists(__DIR__.'/../vendor/autoload.php')
    ? __DIR__.'/../vendor/autoload.php'
    : __DIR__.'/../../../autoload.php'
;
```
More simple?

Issue #349, opened 2017-05-22
Fatal error: Uncaught exception 'Defuse\Crypto\Exception\WrongKeyOrModifiedCiphertextException' with message 'Ciphertext is too short.' 


iyasilias:
I get this error when trying to decrypt a text:

`Fatal error: Uncaught exception 'Defuse\Crypto\Exception\WrongKeyOrModifiedCiphertextException' with message 'Ciphertext is too short.' `

But when I decrypt with simple test file, the encrypted text successfully decrypted.

paragonie-scott:
Are you passing the filename instead of the file contents?

iyasilias:
Sorry my bad. It is actually an error somewhere else (mysql result is null).
Maybe instead of 'Ciphertext is too short', you could  differentiate a message for **null** string.

paragonie-scott:
I created a pull request that throws a `TypeError` (polyfilled by random_compat on PHP < 7) if you pass anything other than a string.

Does your code pass `"NULL"` as a string? If so, that's an oddly specific use case that we can't afford to handle for everyone, because it adds needless complexity and risk for side-channels. If it's `null` (as in an empty value), then #350 will address this sort of confusion.

iyasilias:
Just for record, it turns out the query result is _NULL._

PR #350, opened 2017-05-22
Ensure public API is strictly typed.

paragonie-scott:
Similar to what was suggested in #349.


paragonie-scott:
A quick note about semver: Passing a non-string variable in these contexts would have been an erroneous usage of this library, so throwing a different exception earlier doesn't break BC. This is (conservatively) safe to include in 2.2.0, but arguably fits in a hypothetical 2.1.1 release too.

Issue #371, opened 2017-11-14
Question: why isn't change password implemented?


artfulrobot:
Come across this library for the first time and looks great, thanks.

I have a lot of data that I wish to store encrypted in a database, but over time and at key times I want to change the password (e.g. when staff leave). I don't want to have to decrypt and re-encrypt all the data just to change the password.

I can see that it is possible to change a password by obtaining the decrypted key, then basically repeating most of the code in `createRandomPasswordProtectedKey`.

But as it's something that seems obvious, could be trivially implemented, yet is not implemented, I'm wondering if there's something I've missed? I don't specialise in crypto and am aware that there's a lot I don't know.

paragonie-security:
Relevant: http://archive.hack.lu/2017/hacklu-crypto-api.pdf

I think that would be a nice value-add for v2.2.0

defuse:
Added in #393

PR #373, opened 2017-11-14
Add changePassword method, addressing #371

artfulrobot:


defuse:
Some possible usability pitfalls here are:

- Someone thinks just calling changePassword is enough, and doesn't actually save the key again with the new password (seems likely with this API).
- Someone calls changePassword and saves the new key but leaves old ones lying around, so the old password is still effectively usable.
- anything else?

What do we think about this API in terms of usability? I can imagine another way to do it, which would be to have `changePassword` return a new instance of `KeyProtectedByPassword` and disable or leave the old instance alone. Maybe that would help with the first point, but I'm not sure.

paragonie-security:
Add a `__destruct()` method that throws an exception if `save()` was never invoked.

defuse:
I think I'll bring this in as-is, with documentation and adding a test to make sure the old password doesn't still decrypt the old key. I don't think there is an API design that could prevent people from not deleting old encrypted keys, and if someone forgets to save the new key, then they probably won't have overwritten the old encrypted key, so they won't lose any data (unless they delete/forget the old password).

defuse:
I pulled this into #393.

PR #389, opened 2018-04-22
Misc 2.2 stuff

defuse:


defuse:
There is a bug in psalm that thinks our check for integer overflow in CTR code is a mistake, I reported it upstream: https://github.com/vimeo/psalm/issues/691

PR #405, opened 2018-05-09
Add exceptions when input and output are same in file encrypt or decrypt methods

supertanuki:


defuse:
This is [in the code now](https://github.com/defuse/php-encryption/blob/683b3cf5ca1a30f458a608ea8d8d60c48b69f8ec/src/File.php#L214-L216) (modified so that it works even if the user passes symbolic links to the same file), thank you for the contribution!

PR #408, opened 2018-06-06
Allow random_compat v9.99.99 on PHP 7 projects.

paragonie-scott:
For PHP 7+ software, we have a special tag at `v9.99.99` which is essentially empty. The purpose of this tag is to allow it to be loaded by Composer for PHP 7 software, and then not create conflicting definitions for static analysis tools and/or IDEs' type hinting.

See https://twitter.com/CiPHPerCoder/status/1004230888169852928

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,7 @@
         }
     },
     "require": {
-        "paragonie/random_compat": "~2.0",
+        "paragonie/random_compat": ">= 2",
         "ext-openssl": "*",
         "php":  ">=5.4.0"
     },
--- a/psalm.xml
+++ b/psalm.xml
@@ -7,6 +7,7 @@
     </projectFiles>
     <issueHandlers>
         <DocblockTypeContradiction errorLevel="info" />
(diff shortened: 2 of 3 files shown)


composer.json
```diff
@@ -21,7 +21,7 @@
         }
     },
     "require": {
-        "paragonie/random_compat": "~2.0",
+        "paragonie/random_compat": ">= 2",
```
defuse: Will all versions of random_compat >= 2 have an API that's backwards compatible with the way this library uses it?

paragonie-scott: The API isn't going to change in v3, v4, etc. since it's compatible with PHP 7.

Sephster:
@paragonie-scott is there any timescale on getting this approved? As noted by others, it is causing downstream problems for Laravel Passport and the League's OAuth2 Server at present. Any assistance to expedite this would be much appreciated.

paragonie-scott:
Hopefully that fixes (or outright avoids) the Psalm false positives.

PR #417, opened 2018-09-15
moved to php 5.6 minimum

samnela:
According to the README.md   https://github.com/defuse/php-encryption#php-encryption


NicolasCARPi:
This should be merged imho. There is no point in using electricity to test for unsupported PHP versions :)

PR #421, opened 2018-10-15
Polyfill AES-256-CTR using AES-256-ECB

paragonie-scott:


paragonie-scott:
Tests pass, just needs more code coverage. I'll add unit tests tonight.

defuse:
Thank you so much for this!

src/Core.php
```diff
@@ -36,6 +36,82 @@
...
+     * @return string
+     * @throws Ex\EnvironmentIsBrokenException
+     */
+    public static function polyfillAes256Ctr($plaintext, $key, $nonce)
+    {
+        static $aes256ecbNative = null;
+        if (\is_null($aes256ecbNative)) {
+            $aes256ecbNative = \in_array('aes-256-ecb', \openssl_get_cipher_methods(), true);
+            if (!$aes256ecbNative) {
+                throw new Ex\EnvironmentIsBrokenException(
```
defuse: Suppose we make it here and throw this exception. Now `$aes256ecbNative` is non-null and next time this function gets called we skip over the check, and end up calling `openssl_encrypt` anyway!

src/Core.php
```diff
@@ -36,6 +36,82 @@
...
+        $length = self::ourStrlen($plaintext);
+        /** @var int $numBlocks */
+        $numBlocks = (($length - 1) >> 4) + 1;
+        $stream = '';
+        for ($i = 0; $i < $numBlocks; ++$i) {
+            $stream .= $nonce;
+            $nonce = self::incrementCounter($nonce, 1);
+        }
+        /** @var string $xor */
+        $xor = \openssl_encrypt(
```
defuse: The docs say this can return `false` on "failure." I'm not sure what "failure" means and it seems doubtful that this call could fail but let's check that the result isn't `false` so that there's no chance of XORing the plaintext with "false".

src/Core.php
paragonie-scott: Oh, good catch. I'll move the throw condition down.

src/Core.php
paragonie-scott: Will do.

paragonie-scott:
I'm sending an alternate PR that includes an ECB polyfill using CBC mode. The extra complexity makes me a bit nervous, but "we'll just use AES-256-CBC" is a bad outcome.

Issue #420, opened 2018-10-15
Unsupported cipher method aes-256-ctr


markus-hsk:
At the moment i try to fix an issue on my server:
> Cipher method not supported. This is normally caused by an outdated version of OpenSSL (and/or OpenSSL compiled for FIPS compliance). Please upgrade to a newer version of OpenSSL that supports aes-256-ctr to use this library.

After some searching i found two constants in the Core class: _CIPHER_METHOD_ and _LEGACY_CIPHER_METHOD_. Is it possible to set the cipher method from outside without manipulating the original code? Did not find any possibility to do so.

_I'm using php7.2 on different OS (Ubuntu, Alpine Linux and Windows 10) with v2.2.1 of the defuse/php-encryption package, and LibreSSL 2.7.4_

paragonie-scott:
> Is it possible to set the cipher method from outside without manipulating the original code? Did not find any possibility to do so.

No, that constant is only used in `legacyDecrypt()`.

@defuse I think we really need to do [what I did with CipherSweet](https://github.com/paragonie/ciphersweet/blob/da967be3575b6f18823546f51f75625123d35a83/src/Util.php#L16-L47) and provide an AES-256-CTR polyfill using AES-256-ECB to prevent people from doing anything dangerous.

Of course, the "correct" fix is: Install a newer version of OpenSSL that isn't compiled in FIPS mode so AES-256-CTR is available.

But I'd rather slightly increase this library's attack surface than encourage people to do dangerous things like manipulate the library's code to use an different cipher mode.

markus-hsk:
Sorry, if this might sounds dump, but the installed version of LibreSSL is already the latest stable version 2.7.4. So, there is no possibilty to update and i cannot see any FIPS mode hint.

paragonie-scott:
Weird. LibreSSL removes AES-256-CTR?

paragonie-scott:
Proposed fix in #421 

@markus-hsk Can you please verify that this branch fixes your problem?

paragonie-scott:
If you've never tested a branch before:

`composer require defuse/php-encryption:dev-aes256ctr`

Then run `test.sh`.

markus-hsk:
Hello and good morning,
yeah, thank you. I'll try, but might take some time. I'll come back, then it's implemented and tested.

paragonie-scott:
Can you at least verify for me that `aes-256-ecb` is available?

```php
[...]

PR #422, opened 2018-10-22
AES-256-CTR Enhanced Polyfill

paragonie-scott:
Now based on AES-256-CBC if ECB isn't available. Supersedes the effort in #421, should solve the corner case identified in #420.

`<insert zalgo meme here>`

### DIFF
--- a/src/Core.php
+++ b/src/Core.php
@@ -146,8 +146,9 @@ final class Core
         $start = 0;
         $iv = str_repeat("\0", 16);
         for ($i = 0; $i < $blocks; ++$i) {
+            /** @var string $piece */
             $piece = \openssl_encrypt(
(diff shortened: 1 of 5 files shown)


paragonie-scott:
I think Github's outage earlier today broke the Travis CI integration. :(

defuse:
I'm going to leave this PR open for the time being, since #420 is still open and others might have a similar problem. Before merging this, though, I want to get a better understanding of what environments this is actually needed for, since this is a pretty risky change. How easy/difficult is it to set up an environment that needs this and do we know anything about how common they are?

defuse:
This is the most amazing thing ever. I'm closing it for the obvious reasons, it'll still be here if anyone *really* needs it.

PR #431, opened 2019-01-09
Added testing for PHP 7.3 in .travis.yml

mesour:


peter279k:
The Travis CI build is failed because of dependencies problems.

The error message is as follows:

`Package vimeo/psalm at version dev-master has a PHP requirement incompatibl  
  e with your PHP version (7.0.33)`

I think it should use the stable version and not use the unstable `dev-master` version.

Issue #435, opened 2019-04-11
Fix mbstring.func_overload CI errors


defuse:
https://travis-ci.org/defuse/php-encryption/jobs/513078638

defuse:
Fixed in #437

Issue #436, opened 2019-04-11
Fix Psalm CI warnings and errors 


defuse:
https://travis-ci.org/defuse/php-encryption/jobs/477380503

defuse:
Fixed in #437

Issue #430, opened 2019-01-09
Bump the maintenance lifetime to 2020


defuse:

defuse:
Done in #437

PR #437, opened 2019-04-11
Fix CI errors

defuse:


defuse:
This will close #435, #436, and #430.

PR #441, opened 2019-06-02
Modernize the PHP-CS-Fixer rules

NicolasCARPi:
* Require PHP-CS-Fixer as dev dependency
* Modify .gitignore to ignore php_cs.cache and generated coverage files
* Run the "fix" command on the codebase


NicolasCARPi:
Note that the tests are failing with travis for an unrelated reason.

PR #442, opened 2019-06-02
Add option to speed up the tests. Fix #400

### LINKED ISSUE #400: Speed up the tests
Ideally, the tests should take less than 60 seconds to run. It's probably fine to sacrifice test cases (e.g. encrypting super huge files) to speed up the tests, as long as we're sure it won't lose any line-coverage.

NicolasCARPi:
* Check if first argument of `test.sh` is a file instead of its
existence
* Add "fast" argument to skip test in "@group slow"
* Small fixes to the bash scripts (quotes, syntax, failsafe)
* Add CONTRIBUTING.md file with instructions for tests and syntax fixing

On my 4 GHz CPU, this brings the tests from 109 seconds down to 63
seconds.

### DIFF
--- a/test/unit/CtrModeTest.php
+++ b/test/unit/CtrModeTest.php
@@ -139,7 +139,6 @@ class CtrModeTest extends PHPUnit_Framework_TestCase
 
     /**
      * @expectedException \Defuse\Crypto\Exception\EnvironmentIsBrokenException
-     * @group slow
      */
     public function testIncrementByZero()
     {
@@ -161,7 +160,6 @@ class CtrModeTest extends PHPUnit_Framework_TestCase
     /**
      * @dataProvider allNonZeroByteValuesProvider
      * @expectedException \Defuse\Crypto\Exception\EnvironmentIsBrokenException
-     * @group slow
      */
     public function testIncrementCausingOverflowInFirstByte($lsb)
     {
@@ -173,7 +171,6 @@ class CtrModeTest extends PHPUnit_Framework_TestCase
 
     /**
      * @expectedException \Defuse\Crypto\Exception\EnvironmentIsBrokenException
-     * @group slow
      */
     public function testIncrementWithShortIvLength()
     {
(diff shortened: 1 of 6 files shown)


test/unit/CtrModeTest.php
```diff
@@ -139,6 +139,7 @@ public function testIncrementByNegativeValue()
 
     /**
      * @expectedException \Defuse\Crypto\Exception\EnvironmentIsBrokenException
+     * @group slow
```
defuse: This one shouldn't be slow.

test/unit/CtrModeTest.php
```diff
@@ -160,6 +161,7 @@ public function allNonZeroByteValuesProvider()
     /**
      * @dataProvider allNonZeroByteValuesProvider
      * @expectedException \Defuse\Crypto\Exception\EnvironmentIsBrokenException
+     * @group slow
```
defuse: This one should also be really fast.

test/unit/CtrModeTest.php
```diff
@@ -171,6 +173,7 @@ public function testIncrementCausingOverflowInFirstByte($lsb)
 
     /**
      * @expectedException \Defuse\Crypto\Exception\EnvironmentIsBrokenException
+     * @group slow
```
defuse: This one should be really fast too.

test/unit/CtrModeTest.php
```diff
@@ -182,6 +185,7 @@ public function testIncrementWithShortIvLength()
...
```
defuse: Same

test/unit/CtrModeTest.php
```diff
@@ -193,6 +197,7 @@ public function testIncrementWithLongIvLength()
...
```
defuse: Same

defuse (commented):
I'm not sure about this. Some of the tests marked as slow should be fast and are quite crucial for security. I think it's worth paying the CPU time to run the full test suite. Any tests that encrypt/decrypt super huge files could be marked as slow this way though.

NicolasCARPi:
Hello,

I pushed some commits to address your comments and fix the conflict.

PR #445, opened 2019-08-28
fix #444 added composer require in README.MD

### LINKED ISSUE #444: Add composer command in your README.MD
You should add `composer require defuse/php-encryption` because it's faster than search in the whole doc or even more faster than searching on packagist.
Btw, cool lib.

DevFelixDorn:


defuse:
Thanks! (Included in #446).

PR #447, opened 2019-10-17
Fix Psalm to unbreak other Pull Requests

paragonie-scott:


paragonie-scott:
Every other pull request is breaking because of Psalm versioning.

PR #463, opened 2020-10-20
Fix insecure session recommendation

### LINKED ISSUE #454: In scenario #2, Admin can read session files
In scenario #2 in the tutorial, Admin can get the key by reading the session files. I think, it is need to store an another key as a cookie to encrypt sessions.

defuse:
Fixes #454


defuse:
bloody hell https://travis-ci.org/github/defuse/php-encryption/jobs/737545982

Issue #466, opened 2020-12-15
PHP 8: Invalid opcode 117/2/0. at vendor/defuse/php-encryption/src/Core.php:412


wshafer:
When upgrading to PHP 8 the following error occurs:

```php
Invalid opcode 117/2/0. at vendor/defuse/php-encryption/src/Core.php:412
```

This appears to be a PHP bug and I have submitted a bug upstream: https://bugs.php.net/bug.php?id=80519

Submiting a PR that will correct this until the bug is fixed is upstream

wshafer:
Duplicate of #464

PR #467, opened 2020-12-15
Fix for PHP 8 + Opcache with error "Invalid opcode 117/2/0." 

wshafer:
This PR corrects issue #466 when running PHP 8 + Opcache


wshafer:
Received a note that the upstream bug causing this issue has been fixed and patched upstream and will be released with 8.0.1.  Closing this PR as it will no longer be needed.

PR #484, opened 2021-05-16
Normalise Licence file

exussum12:
Github doesn't pick this up as MIT, made some minor changes for it to be identified correctly


defuse:
Thanks!

PR #504, opened 2023-05-30
Made PHPUnit tests pass in PHP 8.0, PHP 8.1 and PHP 8.2

boenrobot:
Made PHPUnit tests pass in PHP 8.0, PHP 8.1 and PHP 8.2, by supporting PHPUnit 10, PHPUnit 8 and PHPUnit 5, complete with config files, one of which is explicitly picked by phpunit.sh based on the PHPUnit version previously determined based on the PHP version. This required dropping support for PHPUnit 4, but that is OK, as PHPUnit 5 still supports PHP 5.6, that being the earliest PHP version supported by this library.

Also fixed the broken exception message assertions in two of the tests, so that tests now pass.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,10 @@ matrix:
       env: USE_PSALM=1
     - php: "8.0"
       env: USE_PSALM=1
+    - php: "8.1"
+      env: USE_PSALM=1
+    - php: "8.2"
+      env: USE_PSALM=1
     - php: "nightly"
       env: USE_PSALM=1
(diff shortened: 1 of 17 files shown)


defuse:
OH MY GOD YOU ARE A HERO THANK YOU THANK YOU THANK YOU HOW DO I SEND YOU MONEY?

Getting this to work has been on the back of my mind for like a year and was really stressing me out. 

YOU ARE AMAZING THANK YOU!

Issue #500, opened 2022-11-02
New release with the fix for PHP 8 support


slawn:
Now that the fix is in place for the PHP8 - is it possible to get a new release published so that we can grab it from composer? Right now the latest release is from April 2021. Thanks

defuse:
This is currently only blocking on #508.

defuse:
Looking at the [diff between the last release and master](https://github.com/defuse/php-encryption/compare/v2.3.1...master), there's no breaking changes, so the next version number will be v2.4.0.

defuse:
Rather than build and test the phar with all versions of PHP, I'm only going to support PHP 8.1+ for phar builds and tests. Otherwise I have to pull down older versions of box, depending on the PHP version, which is arguably unnecessary complexity. Barring any future security vulnerabilities, users of old PHP versions that rely on the phar can use the older releases of this library. Remember to mention this in the next release's notes.

defuse:
v2.4.0 has been released!

PR #511, opened 2023-06-16
Re-enable phar building and testing on CI

defuse:


defuse:
https://github.com/defuse/php-encryption/issues/500#issuecomment-1596471142

PR #524, opened 2024-01-02
composer.json: declare ext-hash dependency

staabm:
it seems its required for the package to work in https://github.com/defuse/php-encryption/blob/9b77beb6e1cd997aecfd2392849db5afab128b5e/src/Crypto.php#L282


cristianbuja (commented):
Sorry bad comment

